PR #1, opened 2016-04-04
Adding Rakefile

joesoftware:
Lots of stuff to fill in, but this is a start.


awong-dev:
LGTM.

Huge PR that I would normally get ornery about, but since it's just committing what's mostly data files and some modified sample code, let's get it in and we can start iterating.

PR #3, opened 2016-04-13
WIP: move hca-api closer to being useful as a REST service.

joesoftware:


awong-dev:
@joesoftware what's the status on this?  Should I be reviewing?

awong-dev:
LGTM  reviewed it anyway.

We're sorta scattering test data files randomly around. We'll want to clean that up soon, but for now let's just check it in.

PR #5, opened 2016-04-26
Migrate client from vets website

awong-dev:
This is using git subtree to split out the old source from vets-website.

In vets-website, the HCA code was mostly located in 2 directories: _health-care/_js and spec/javascript. On the vets-website repo, a synthetic history was created for both those directories using 

```
git subtree split -P _health-care/_js -b main-js-only
git subtree split -P spec/javascripts -b test-js-only
```

This created 2 branches where the respective prefix subdirectories were moved to the "root" of the repository and a new synthetic commit history was created for each that pruned out any changes to files outside these directories.  After that, in the healthcare-application repo, `git subtree add` was used to graft those two branches onto this repo:

```
git subtree add --prefix src/server ../vets-website main-js-only
git subtree add --prefix test/server ../vets-website test-js-only
```


awong-dev:
Not waiting for a review on this. ha ha ha ha ha ha.

PR #6, opened 2016-04-26
Rename src/server to src/client and kill old code.

awong-dev:


awong-dev:
Not asking for a review.

PR #7, opened 2016-04-27
Make it run

awong-dev:


alexose:
The linter is tripping over some extra spaces in `src/server.js`.  Quick fix, but I don't have write access yet.  Ops pls?

PR #8, opened 2016-04-27
Set up Karma to run Unittests.

awong-dev:
Also fixes all the tests to run in this environment.


awong-dev:
Going to preemptively merge. Most of this is copied from vets-website's config so it's not really "new" code.  Not having it in though is blocking others from being able to write tests, etc.

PR #9, opened 2016-04-28
Add mapper function for posting to back-end.

awong-dev:
Gigantic check-in.

Still needs fixing and several items (mentioned in the initial commented
form store section) require input from ES team.

### DIFF
--- a/src/client/utils/mapper.js
+++ b/src/client/utils/mapper.js
@@ -1,0 +1,519 @@
+/*
+const blankChildInformation = {
+  childBecameDependent: { // not found in ES
+    month: null,
+    day: null,
+    year: null
+  },
+  childEducationExpenses: null, // not found in ES
+};
+
+const blankProvider = {
+  // none of these fields are asked in the form.
+  insuranceAddress: null,
+  insuranceCity: null,
+  insuranceCountry: null,
+  insuranceState: null,
+  insuranceZipcode: null,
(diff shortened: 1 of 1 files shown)


alexose:
👍 

This looks totally reasonable at first glance, but it's hard to know whether or not it works without something that mocks the external API.

awong-dev:
@alexose This is @joesoftware 's PR.  I would work with him.. In general we're following the rule though that if someone LGTM's a PR, they should merge it.

However,I don't think Joe actually meant for this to be reviewed quite yet...

alexose:
Okay, I'll hold off for now!

src/client/utils/mapper.js
```diff
@@ -0,0 +1,519 @@
...
+      return 'W';
+    case 'Divorced':
+      return 'D';
+    default:
+      return 'U';
+  }
+}
+
+function makeSpouse(nameAndGeneralInformation, spouseInformation) {
+  if (nameAndGeneralInformation.maritalStatus !== 'Never Married') {
```
robbiethegeek: why not do 
if (nameAndGeneralInformation.maritalStatus === 'Never Married) {
  return undefined; // also why undefined? is there a reason?
}
return {....

joesoftware:
New form store is in flux, but this is being merged so someone can pickup the work.

PR #10, opened 2016-04-28
Remove null from model. Replace with "field."

awong-dev:
Sets the stage for doing validations correctly.


akainic:
Made some fixes, validations should be mostly working now! Only thing that is not working correctly is updates to radio button questions. They are overriding the object `makeField` and are instead setting the value directly. E.g., instead of `isVaServiceConnected: {value: 'Y', dirty: true}`, when an answer is selected the store becomes `isVaServiceConnected: 'Y'`. Not sure why this is happening, but I think this is something we can fix in a subsequent PR. This one probably should be merged in soon. @awong-dev if you want to take another look before we merge that would be great!

awong-dev:
LGTM. Thanks for fixing all that.  Let's get this in.

PR #14, opened 2016-04-29
Fix react devtools loading within an iframe.

awong-dev:
Allows use devtools with http://localhost:3000/webpack-dev-server/ to
get all the hot-reload goodness.


crwallace:
LGTM!

PR #15, opened 2016-05-02
Change required indication

akainic:
From user testing we found that some users had trouble understanding what the "Required" span was indicating. We are instead going to use asterisks to indicate required fields. 

Screenshot (ignore the condensed column of questions, something is messed up with our styles):
screenshot 2016-05-02 10 09 50

@maryannbrody @gnakm Would love to get your input as well on this method of indicating required fields. Should we may the asterisk red?

### DIFF
--- a/src/client/components/form-elements/DateInput.jsx
+++ b/src/client/components/form-elements/DateInput.jsx
@@ -58,9 +58,18 @@ class DateInput extends React.Component {
           validateIfDirtyDate(day, month, year, isValidDate);
     }
 
+    // Calculate required.
+    let requiredSpan = undefined;
+    if (this.props.required) {
+      requiredSpan = <span className="hca-required-span">*</span>;
+    }
+
     return (
       <div>
-        <label>{this.props.label ? this.props.label : 'Date of Birth'}</label>
+        <label>
+          {this.props.label ? this.props.label : 'Date of Birth'}
+          {requiredSpan}
+        </label>
         <span className="usa-form-hint usa-datefield-hint" id="dobHint">For example: Apr 28 1986</span>
         <div className={isValid ? undefined : 'usa-input-error'}>
           <div className="usa-date-of-birth row">
--- a/assets/css/hca.scss
+++ b/assets/css/hca.scss
@@ -102,6 +102,11 @@ select {
   }
 }
 
+.hca-required-span {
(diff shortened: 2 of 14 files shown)


gnakm:
@akainic I totally think we should make the asterisks red because it will be easier to scan.

crwallace:
Also, do we think in addition the text should say something like, "(*) indicates that this question must be ANSWERED to continue." To help people that are confused by the meaning of "required"? @maryannbrody thoughts from a literacy standpoint?

akainic:
@crwallace @gnakm What do you think of this in terms of how the red asterisk looks?
screen shot 2016-05-02 at 11 21 09 am

gnakm:
LGTM!

crwallace:
@akainic That looks awesome!

maryannbrody:
@akainic @crwallace @gnakm -- visually, this looks great!

From a plain language perspective.. this is pretty common web convention, so we can probably get away with it.. but... here are some more plain language options
- "You must fill in each piece of information with a (*) star next to it."
- "To be able to send this form, fill in each piece of information with a (*) star next to it."

akainic:
@maryannbrody Maybe we leave it as is and user test again to make sure the language is clear?

src/client/components/form-elements/DateInput.jsx
```diff
@@ -58,9 +58,18 @@ class DateInput extends React.Component {
           validateIfDirtyDate(day, month, year, isValidDate);
     }
 
+    // Calculate required.
+    let requiredSpan = undefined;
+    if (this.props.required) {
+      requiredSpan = <span className="hca-required-span"> *</span>;
+    }
+
```
crwallace: I think this looks good for now but I wonder if, for future forms, we should have a <requiredAsterisk/>. Maybe for component library down the road.

crwallace:
LGTM!

src/client/components/form-elements/DateInput.jsx
akainic: Yes, totally agree. Maybe instead of a `' '` we add margin around the span?

crwallace: Yes and when we write documentation we can use code examples to show people how to use it.

PR #16, opened 2016-05-02
Fix the main CSS issues

akainic:
A few of the CSS issues we were seeing:
1. Not all the right classes were being applied in the header and footer
2. Changes were made directly to the USWDS files in vets-website, so I had to add custom styles here to match those styles that will override the USWDS files
3. Svg down arrow in selects was not rendering. Instead reordered styles so that png gets rendered, which is working. Probably should figure out why svgs weren't rendering in the first place, though.

### DIFF
--- a/assets/css/hca.scss
+++ b/assets/css/hca.scss
@@ -3,9 +3,21 @@ body .row {
   max-width: 62.5em;
 }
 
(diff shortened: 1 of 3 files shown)


awong-dev:
LGTM

Reminds me to set up a linter for our CSS.

PR #18, opened 2016-05-03
Add Dockerfile and build script for deploy.

awong-dev:
Adds a build script that creates a Docker image suitable for deploying.


Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Use Alpine-node as a base image because it is much smaller.
+#
+# Base image size for mhart/alpine-node:4 is 13 MB
+# Base image for standard node:4 is 253 MB.
+#
+# Our production deploy does not have any C/C++ library dependencies yet
+# so this is still easy to do.
+FROM mhart/alpine-node:4
```
plusjeff: Most of the image makes sense to me, but I really can't figure out what this command does:
https://github.com/mhart/alpine-node/blob/01ed9039d53628391fe4ccae4ae9dfd32797d701/Dockerfile#L30
Looks like it makes sure that the executable can't do something to memory, but not following 100%. It's probably fine, just curious why this image was chosen.

jeffmaher:
Could you add a quick bit of documentation to the README explaining:
- How to start the container (i.e. what the `docker run...` command looks like)
- How someone can hit the server once they've run it

Dockerfile
plusjeff: Also thinking about this more. It might be good to use a CentOS derived image. We've been flipping back and forth between distros, and CentOS-based ones are what all the current Ansible scripts are based on so that we have consistency on what to expect from both host and container OSes.

awong-dev: Wow you're thorough. :)  paxctl is command that modifies the binary to set some flags that are examined if your kernel is built with grsecurity hardening.  grsecurity is a cool suite of tools that allow for much stricter memory protections. For example, you can mark your stack and heap as non-executable completely eliminating the standard buffer-overflow vulnerability for code injection.  However, node.js does JIT so it needs to be able to write executable code. This command is marking the node executable as needing to disable those memory protections.

As for why this image was chosen, I originally thought it'd be best to use CentOS or something...but my understanding has changed recently.  Eric Maland pointed out that docker is conceptually one task. The fact that the container has to have an OS to bootstrap that process is sorta incidental. IIRC, the ideal docker setup is actually a 100% statically linked binary with almost no supporting filesystem.

Thus, ansible has no place in docker. Instead, you would create one docker image per functional unit (making up a term here) and then use docker-compose to weave them together into what would be the equivalent of a playbook.

In this world, all standard distros are too heavy weight. Alpine linux is a super stripped down linux that's meant to just run one binary. Thus using alpine-node (or alpine based nginx) makes the most sense to me for a container OS....

awong-dev:
@plusjeff comments addressed. Back at ya.

Dockerfile
plusjeff: Yep - totally understand, no Ansible involved in the Docker setup, but there is for getting the Docker image to the host and any apps not using Docker (or is that Docker compose...don't know about that piece of the ecosystem?). I'm thinking more along the lines of the commands in the Docker file will be easier to learn if the host and container OSes are the same (i.e. files go in the same kinds of places). In other words, it's more of a people thing; folks don't have to learn the little nuances of each distro for setting things up.

However, in this case, I suspect that the performance gains of having a stripped down Linux are more compelling, so I'm cool with it being Alpine. Just curious how you stumbled upon this Alpine versus one of the ones marked as _Official_ (such as https://hub.docker.com/_/alpine/ ).

jeffmaher:
@awong-dev : Just one more question, otherwise the rest LGTM. Will bounce back to you for that.

Dockerfile
awong-dev: I actually don't believe there should be any code sharing between the Docker deploy and Ansiable.

If we were thinking of the docker deploys as being fully formed sub-operating systems that run more like a light-weight VM, then I agree. However, that sounds more and more like an anti-pattern. For example, I wouldn't ever want to use upstart to manage my process inside the docker container. Instead, I would want my process to be the equivalent of the OS init, and then have something monitor the `docker run` call and restart there.

As for how I stumbled on this image....I didn't actually do much.... just Google searched and checked to see if it was commonly used. Note that the mhart/alpine-node image is a layer on top of the official alpine linux install which just adds node. See https://github.com/mhart/alpine-node/blob/master/Dockerfile#L1.

awong-dev:
@plusjeff responded to comment. Back at ya!

awong-dev:
I'm going to merge this for now.

PR #20, opened 2016-05-05
Add react-scroll and use that to handle scrolling

akainic:
Fix how we were handling scrolling before by using `react-scroll` and scrolling to an element at the top of the page. I think this is better than searching the DOM for specific nodes and scrolling to those using plain old JS, but it also gives us less control over where we're scrolling to. As far as I can tell there's no way to scroll to a particular error field using `react-scroll`. Instead, I'm always scrolling to the top of the form. I think this is okay because 1) scrolling to a particular error field wasn't working before anyway (wouldn't scroll until the second time you click on the button because of when the function is being called in relation to when the error classes are added to the page), and 2) in our new version of the form with fewer questions and shorter answers, you should be able to see all error fields if we scroll to the top anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "phantomjs-prebuilt": "^2.1.4",
     "react-addons-test-utils": "^15.0.1",
     "react-redux": "^4.4.5",
+    "react-scroll": "^1.0.15",
     "react-transform-catch-errors": "^1.0.2",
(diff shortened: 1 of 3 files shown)


crwallace:
This is awesome and I agree with you about the behavior once we have smaller sections. LGTM!

PR #22, opened 2016-05-06
Skelton of new node.js API server for HCA

robbiethegeek:
Structure of application.
Vanilla Express.
Testing in place.
Code Coverage in place.
Linting in place.

### DIFF
--- a/hca-api-server/package.json
+++ b/hca-api-server/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "hca-api-server",
+  "description": "API server for the Health Care Application.",
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "start": "node ./src/server",
+    "lint": "eslint ./src/**/*.js ./specs/**/*.js",
+    "test": "istanbul cover _mocha -- ./specs --recursive",
+    "cc:display": "open coverage/lcov-report/*.html"
+  },
+  "dependencies": {
+    "body-parser": "~1.12.0",
+    "cookie-parser": "~1.3.4",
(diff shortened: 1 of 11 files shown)


hca-api-server/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "dependencies": {
+    "body-parser": "~1.12.0",
+    "cookie-parser": "~1.3.4",
+    "debug": "~2.1.1",
+    "express": "~4.12.2",
+    "morgan": "~1.5.1"
+  },
+  "devDependencies": {
+    "eslint": "^2.9.0",
+    "eslint-config-airbnb-base": "^2.0.0",
```
alexose: Cool!  How did you decide on the airbnb linting rules?

hca-api-server/src/server.js
```diff
@@ -0,0 +1,5 @@
+const app = require('./app');
+
+app.listen(3000, () => {
```
alexose: Might be nice to use https://www.npmjs.com/package/portfinder.  One less thing that can go wrong in the dev environment.

alexose:
Looks great!  Nothing controversial in here, I think!

The complexity shows up once we start plugging in real VA endpoints.  It might be worth talking about queues, persistence, etc.

I'm also eager to figure out how we should test external endpoints...  My first thought is that we should stub everything out (with https://www.npmjs.com/package/sinon) so that tests will pass locally, then write another layer of integration tests that are designed to run on dev/staging/prod.

Of course, none of this matters yet.  Just wanted to get some thoughts on paper.

👍

PR #25, opened 2016-05-09
Update store

akainic:
This PR is mainly an attempt to change our store to not be dependent on the structure of the form (remove fields from being nested within sections and instead fields are all at the same level). The following main changes were made to accomplish this:
1. Passing the entire veteran data store to each section component
2. Changing the UI store to capture the fields that are in each section (this is needed for the initialization function that dirties all the fields in a given section, since the veteran store no longer contains section logic)

I am not confident that 2 above is the right solution to the problem, but it was the best I could come up with. Would love to get feedback on this.

This PR is a major change to how our application processes data, so would love if @alexose could look at the store from an API perspective and if @awong-dev could look see if the approach makes sense.

### DIFF
--- a/src/client/reducers/veteran/index.js
+++ b/src/client/reducers/veteran/index.js
@@ -3,171 +3,140 @@ import lodashDeep from 'lodash-deep';
 
 import { ENSURE_FIELDS_INITIALIZED, VETERAN_FIELD_UPDATE, UPDATE_SPOUSE_ADDRESS, CREATE_CHILD_INCOME_FIELDS } from '../../actions';
 import { makeField, dirtyAllFields } from '../fields';
-import { pathToData } from '../../store';
 
 // Add deep object manipulation routines to lodash.
 _.mixin(lodashDeep);
 
-// TODO(awong): This structure should reflect a logical data model for a veteran. Currently it
-// mirrors the UI stricture too much.
-
 // TODO: Remove providers and children if checkbox within section is unchecked
 const blankVeteran = {
-  nameAndGeneralInformation: {
-    fullName: {
-      first: makeField(''),
-      middle: makeField(''),
-      last: makeField(''),
-      suffix: makeField(''),
-    },
-    mothersMaidenName: makeField(''),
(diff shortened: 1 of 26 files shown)


robbiethegeek:
Wow this is a huge refactor and it looks great from a first pass I will try to dig in deeper shortly.

src/client/reducers/veteran/index.js
```diff
@@ -3,171 +3,140 @@ import lodashDeep from 'lodash-deep';
...
+  wantsInitialVaContact: false,
+
+  isSpanishHispanicLatino: false,
+  isAmericanIndianOrAlaskanNative: false,
+  isBlackOrAfricanAmerican: false,
+  isNativeHawaiianOrOtherPacificIslander: false,
+  isAsian: false,
+  isWhite: false,
+
+  address: {
```
crwallace: I think the way that @joesoftware is setting up the backend, everything that is about a veteran is prefixed with "veteran". So `address` would be `veteranAddress` and `county` would be `veteranCounty`. This way we can differentiate between these fields for a veteran, a spouse or a child.

akainic: Yep, that makes sense. I'll make those changes.

alexose:
These look like good, sound changes.  The big wins:
- The `veteran` object is simplified/flattened
- Components now have access to the whole `veteran` object instead of just a subsection

If it's less boilerplate, I'm all for it.  On the other hand, I see opportunities to represent this data in a clearer way.  (For example, why not store dates as timestamps or UTC strings?).

Anyway, looks good.  The less tightly coupled the better!  👍

akainic:
@alexose Yep, totally agree about some of the data weirdness. We started by keeping the formats of what the enrollment system accepts, but we could store something different in the store and then adjust it in mapper.js to match what the enrollment system accepts. Let's consider that for the next PR if that's okay. Thanks for reviewing this!

akainic:
@awong-dev Going to merge this for now so it's not holding anything else up, but if you want to review and add comments post-merge that we can fix later that would be great!

PR #28, opened 2016-05-10
Adds autoComplete to form.

crwallace:
This PR adds the `autoComplete` attribute to the `form` element to allow users to autofill when supported by a browser.

### DIFF
--- a/src/client/components/HealthCareApp.jsx
+++ b/src/client/components/HealthCareApp.jsx
@@ -149,7 +149,7 @@ class HealthCareApp extends React.Component {
         <div className="medium-8 columns">
           <div className="progress-box">
           {/* TODO: Change action to reflect actual action for form submission. */}
-            <form className="form-panel" autoComplete="on">
+            <form className="form-panel">
               {children}
               {buttons}
             </form>
--- a/src/client/components/HealthCareApp.jsx
+++ b/src/client/components/HealthCareApp.jsx
@@ -148,10 +148,11 @@ class HealthCareApp extends React.Component {
         </div>
         <div className="medium-8 columns">
           <div className="progress-box">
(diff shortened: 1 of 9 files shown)


src/client/components/form-elements/ErrorableSelect.jsx
```diff
@@ -86,6 +88,8 @@ class ErrorableSelect extends React.Component {
 
 ErrorableSelect.propTypes = {
   errorMessage: React.PropTypes.string,
+  name: React.PropTypes.string,
+  autocomplete: React.PropTypes.string,
```
akainic: Sorry this is really tedious, but could we also add these props in the description at the top of the file in the comments? For all the files in this PR you've touched?

src/client/components/HealthCareApp.jsx
```diff
@@ -148,10 +148,11 @@ class HealthCareApp extends React.Component {
         </div>
         <div className="medium-8 columns">
           <div className="progress-box">
-            <div className="form-panel">
+          {/* TODO: Change action to reflect actual action for form submission. */}
+            <form className="form-panel" autoComplete="on">
```
akainic: Do we need to specify `autocomplete="on"` for this to work? I think `on` is the default.

akainic:
LGTM with a couple nits!

src/client/components/form-elements/ErrorableSelect.jsx
crwallace: Will do!

src/client/components/HealthCareApp.jsx
crwallace: I thought so too but for some reason it was not working at first. I will remove and test this.

PR #29, opened 2016-05-10
Restyle navigation buttons

akainic:
New button styles based on outcomes of SF trip. 

screen shot 2016-05-10 at 10 52 12 am


crwallace:
This LGreatTM!

PR #30, opened 2016-05-10
Change field names in validations to match store

akainic:
These were not updated when I updated the store.


robbiethegeek:
👍  LGTM

PR #35, opened 2016-05-12
Add icons to secondary action buttons and change styling

akainic:
Change styles of secondary action buttons to be outline style except for "Save" button. Also added icons to differentiate and reinforce the actions. 

Question to group: Do we want to hide the "Add" button until "Save" is clicked?

screen shot 2016-05-11 at 9 03 16 pm


crwallace:
LGTM!

PR #37, opened 2016-05-12
POST behavior for submit button

alexose:
After spending some time reading about Redux middleware, I decided to go the most straightforward route possible.

A few things:
- This is pointing to a dummy endpoint that doesn't do anything (`/hca`).  Let's change that as soon as we can.
- It would be nice to scrape all of the `isDirty` references out of the JSON.  Should be a lodash one-liner.
- The actions it dispatches aren't hooked up to anything yet.


alexose:
y u no run npm install travis??

awong-dev:
travis runs npm install, but the shirinkwrap overode your package.json update.  You need to run `npm shrinkwrap --dev` to snapshot everything.

src/client/components/HealthCareApp.jsx
```diff
@@ -74,10 +76,38 @@ class HealthCareApp extends React.Component {
...
+    const veteran = store.getState().veteran;
+
+    store.dispatch(updateSubmissionStatus('submitPending'));
+    store.dispatch(updateCompletedStatus(path));
+
+    // Pretty-print the data that we're about to POST
+    console.log(JSON.stringify(veteran, null, 2));
+
+    // POST data to endpoint
+    fetch('/hca/', {
```
awong-dev: This is already here! (sorta).  Try `/v1/api/submit`.

However if you look at:

https://github.com/department-of-veterans-affairs/healthcare-application/blob/master/hca-api-stub/voa-rest.js#L50

you'll notice we have the wrong verb. Oops.

src/client/components/HealthCareApp.jsx
```diff
@@ -74,10 +76,38 @@ class HealthCareApp extends React.Component {
...
+        'Content-Type': 'application/json'
+      },
+      timeout: 10000, // 10 seconds
+      body: JSON.stringify(veteran)
+    }).then(response => {
+      if (!response.ok) {
+        throw new Error(response.statusText);
+      }
+      return response;
+    }).then(response => {
```
awong-dev: Why chain so much in this promise?  There's nothing async happening after the first response...  I feel like you could just call `store.dispatch(updateSubmissionStatus('submitFailed', error));` in line 102 and `store.dispatch(updateSubmissionStatus('submitSucceeded', response.json()));` instead of returning response in 104.

alexose: This way I can wrap up both connection errors and non-ok responses (as per https://www.tjvantoll.com/2015/09/13/fetch-and-errors/).  This is my first attempt at using the fetch API, though, so I might be off my rocker...

awong-dev: Oh interesting.... that makes sense, though I think you can still remove one of the then() blocks.

package.json
```diff
@@ -62,6 +62,7 @@
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "^1.7.0",
+    "less": "^2.7.1",
```
robbiethegeek: never have used this module and doesn't seem to be used in the PR.

alexose: This was an unrelated issue with npm shrinkwrap.  It wouldn't work because of an unresolved peer dependency with less (and also this: https://github.com/npm/npm/issues/6565).

In npm install, it's just a warning, but in shrinkwrap it's an error.  Hence the need to explicitly resolve it.  If there's a better way around this issue, I'd love to know...

awong-dev: FYI... less is needed by the sass compiler in wrbpack iirc.

robbiethegeek: What version of node and npm are you using? I think this is something that was resolved in later versions of npm.

alexose: 3.8.3.  Yeah, it surprised me that it wasn't working...

awong-dev:
LGTM.

PR #42, opened 2016-05-12
Add ES elements and validation info.

awong-dev:
Adds the actual official VOA Data Elements and Validations xlsx file.

Also in this CL, the the relevant columns from the xlsx file for both
the "VOA Data Elements" tab as well as the "Data Validations" tab have
been combined into a CSV format and inserted as comments above the
relevant functions inside enrollment-system.js.

This should allow parallel implementation of the different form
sections.


awong-dev:
PTAL

robbiethegeek:
This looks good, question about using the function keyword rather then the fat arrow syntax.

`function veteranToEnrollmentDeterminationInfo(veteran) {`
vs.
`const veteranToEnrollmentDeterminationInfo = (veteran) => { //no need for the parens but I find pretty unreadable`

awong-dev:
We clarified on slack, but putting it here for completeness.

Regarding function keyword, the rule thus far has been using function keyword for function statements and fat-arrow function for function expressions.

If I recall correctly, function statements are technically only legal at the top level and not nested in blocks (even though most browser and runtimes support them there) so really function expressions mostly come into play with callbacks. In that case, fat-arrow's lexical this semantics are superior. The linter is actually configured to barf on function expressions.

These aren't set in stone, but that's the current convention.

PR #44, opened 2016-05-12
UI clean up for ReviewAndSubmitSection

crwallace:


crwallace:
@akainic I would also love your feed back on the `{childrenIncomeInput}` component. It does not appear to be rendering when you add a child. I don't think I understand how `<FixedTable/>` works but I think something is getting passed to that incorrectly.

src/client/components/insurance-information/AdditionalInformationSection.jsx
```diff
@@ -14,8 +14,26 @@ import { validateIfDirty, isNotBlank } from '../../utils/validations';
  * `reviewSection` - Boolean. Hides components that are only needed for ReviewAndSubmitSection.
  */
 class AdditionalInformationSection extends React.Component {
+  constructor() {
+    super();
+    this.displayLabel = this.displayLabel.bind(this);
+  }
+
+  displayLabel(fullList, userValue) {
```
akainic: Can we rename this to be more specific? Like `displayFacilityLabel`?

src/client/components/military-service/ServiceInformationSection.jsx
```diff
@@ -12,15 +12,33 @@ import { veteranUpdateField } from '../../actions';
  * `reviewSection` - Boolean. Hides components that are only needed for ReviewAndSubmitSection.
  */
 class ServiceInformationSection extends React.Component {
+  constructor() {
+    super();
+    this.displayLabel = this.displayLabel.bind(this);
+  }
+
+  displayLabel(fullList, userValue) {
```
akainic: Ah, this is exactly the same as the other function. Perhaps we can export it to a shared utils file, perhaps even in `calculated.js` and then just import it in the files you need?

akainic:
The review panels looked a little compressed because of the `.form-panel` class that sets a max-width. Do you think we can change that for the review page so that it takes up the full-width of the `progress-box`?

akainic:
Also the buttons are a little weird in the review and submit page
screen shot 2016-05-12 at 7 33 13 pm

akainic:
@crwallace I see what you mean about the child income fields not being updated if an additional child is added in the review page. I think this is because the action that creates the additional children income fields happens when annual income renders, not when children are added in child information section. A way to fix this for the review page might be to collapse all sections and only render one at a time, like we talked about originally doing.

src/client/components/military-service/ServiceInformationSection.jsx
crwallace: I didn't originally do this because `calculated.js` seemed like it was supposed to be for things related to state but I added it with a comment.

crwallace:
@akainic Regarding the buttons. Because of the specified styles, we can either add `!important`s (blec) or add the styles for the "add" and "edit" buttons inside of `.hca-review-panel`. But doing that will be duplicate code.

akainic:
@crwallace Yep, I see what you mean about the buttons. The `accordion` styling is overriding the button styling we use elsewhere. We can table that issue for now and discuss this afternoon what these buttons should look like, and then fix that in a follow-up PR.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -52,12 +53,12 @@ class ReviewCollapsiblePanel extends React.Component {
...
       editButton = (<button
           className="edit-btn"
-          onClick={this.handleEdit}>Edit</button>
+          onClick={this.handleEdit}>Edit Section</button>
       );
     } else {
       panelAction = (<button
           className="usa-button-outline"
-          onClick={this.handleSave}>Save</button>
+          onClick={this.handleSave}>Update Section</button>
```
akainic: I like this language better!

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -68,22 +69,40 @@ class ReviewCollapsiblePanel extends React.Component {
...
+    const sectionIndexes = allSections.indexOf(currentPath);
+    const prevPath = allSections[sectionIndexes - 1];
+
+
+    if (sectionsVerified) {
+      hiddenSection = (<div></div>);
+    } else {
+      if (this.props.uiData.sections[prevPath].verified || currentPath === '/veteran-information/personal-information') {
+        hiddenSection = (
+          <div id={`collapsible-${this.id}`} aria-hidden="false" className="usa-accordion-content">
```
akainic: Do we still need the `aria-hidden` attribute now that we're not hiding this anymore?

akainic:
LGTM with one question!

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
crwallace: Correct we do not. I will remove it.

PR #46, opened 2016-05-13
Update validations

akainic:
Updated validations and added required fields that are required in VOA, but were not indicated to be required here. There are still some checkbox fields in our form that are required in VOA; need to decide whether or not to change those to radio buttons.

### DIFF
--- a/src/client/utils/validations.js
+++ b/src/client/utils/validations.js
@@ -73,6 +73,10 @@ function isValidDate(day, month, year) {
     return false;
   }
 
+  if (Number(year) < 1900) {
+    return false;
+  }
+
   return date.getDate() === Number(day) &&
     date.getMonth() === adjustedMonth &&
     date.getFullYear() === Number(year);
@@ -91,7 +95,7 @@ function isValidMonetaryValue(value) {
 
 // TODO: look into validation libraries (npm "validator")
 function isValidPhone(value) {
-  return /^\d{3}-\d{3}-\d{4}$/.test(value);
+  return /^\d{10}$/.test(value);
 }
 
 function isValidEmail(value) {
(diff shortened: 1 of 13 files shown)


src/client/utils/validations.js
```diff
@@ -180,14 +212,18 @@ function isValidVeteranAddress(data) {
       isValidField(isValidPhone, data.mobilePhone);
 }
 
+function isValidFinancialDisclosure(data) {
+  return isNotBlank(data.provideFinancialInfo.value);
+}
+
 function isValidSpouseInformation(data) {
```
crwallace: For the `isValidSpouseInformation` validation, do we want to make the spouse fields only validate `ifNotBlank` when the user selects `married`?

akainic: Yes, definitely, these validations should be updated. I was waiting to do anything with the spouse section because the conditional logic isn't in place yet. I'll add a TODO for now here.

src/client/utils/validations.js
```diff
@@ -180,14 +212,18 @@ function isValidVeteranAddress(data) {
...
       isValidField(isValidSSN, data.spouseSocialSecurityNumber) &&
       (isBlankDateField(data.spouseDateOfBirth) || isValidDateField(data.spouseDateOfBirth)) &&
       (isBlankDateField(data.dateOfMarriage) || isValidDateField(data.dateOfMarriage)) &&
       (isBlankAddressField(data.spouseAddress) || isValidAddressField(data.spouseAddress)) &&
       isValidField(isValidPhone, data.spousePhone) &&
-      isNotBlank(data.maritalStatus);
+      isNotBlank(data.maritalStatus.value);
 }
 
 function isValidChildInformationField(child) {
```
crwallace: Right now the form will still validate even if `DateChildBecameDependent` is before `ChildDateOfBirth`. Maybe this should be equal to or less than?

src/client/utils/validations.js
```diff
@@ -245,19 +281,19 @@ function isValidDeductibleExpenses(data) {
...
+function isValidVAFacility(data) {
   return validateIfDirty(data.facilityState, isNotBlank) &&
     validateIfDirty(data.vaMedicalFacility, isNotBlank);
 }
 
 function isValidMedicareMedicaid(data) {
-  return isBlankDateField(data.medicarePartAEffectiveDate) ||
-    isValidDateField(data.medicarePartAEffectiveDate);
+  return validateIfDirty(data.isMedicaidEligible, isNotBlank) &&
+    validateIfDirty(data.isEnrolledMedicarePartA, isNotBlank) &&
```
crwallace: Does `isEnrolledMedicarePartA` need to be required if my answer to `isMedicaidEligible` is "no"?

src/client/utils/validations.js
```diff
@@ -245,19 +281,19 @@ function isValidDeductibleExpenses(data) {
...
+    (isBlankDateField(data.medicarePartAEffectiveDate) ||
+    isValidDateField(data.medicarePartAEffectiveDate));
 }
 
 function isValidGeneralInsurance(data) {
```
crwallace: Looks like I am still able to save my `providers` even though the form is in valid.

crwallace:
@akainic this looks awesome! You are awesome for tackling all of this work! Just a couple of questions but everything else looks great to me.

src/client/utils/validations.js
akainic: Good catch! I will fix this.

src/client/utils/validations.js
akainic: I think they are independent of each other. The original form has both as required fields with no conditional logic, so I just followed that.

src/client/utils/validations.js
akainic: Uh oh! I will look into this.

PR #47, opened 2016-05-13
Add debug components for easier testing.

awong-dev:
Adds PopulateVeteranButton button which prepopulates a bunch of fields
to make form testing way easier.

Adds a RoutesDropdown that allows navigating to subsections much faster.

### DIFF
--- a/src/client/components/HealthCareApp.jsx
+++ b/src/client/components/HealthCareApp.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import Scroll from 'react-scroll';
-import { hashHistory } from 'react-router';
+import _ from 'lodash';
 
 import fetch from 'isomorphic-fetch';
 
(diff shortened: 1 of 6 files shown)


src/client/components/HealthCareApp.jsx
```diff
@@ -173,19 +179,36 @@ class HealthCareApp extends React.Component {
         </div>
       );
     }
+    let devPanel = undefined;
+    if (__DEV__) {
+      const queryParams = _.fromPairs(
+        window.location.search.substring(1).split('&').map((v) => { return v.split('='); }));
```
robbiethegeek: I have never done this, but it looks like react router provides the query params:
Here is an example from here : https://github.com/reactjs/react-router/blob/8707d05b00273590e4d054771b8ae7ae9ad312e3/examples/query-params/app.js
    let { query } = this.props.location
    let age = query && query.showAge ? '33' : ''

crwallace:
@awong-dev this looks good but how do I run it with `dev` specified so that I can see that it works?

crwallace:
Got it running and LGTM!

src/client/components/HealthCareApp.jsx
awong-dev: Nice!  I was looking for something like that but couldn't find it.  I'll fix in a follow-up PR.

PR #49, opened 2016-05-13
Add request logging. Redo shrinkwrap w/ npm 3.

awong-dev:


alexose:
I've never used morgan before, but it seems fine to me.

👍

awong-dev:
@alexose morgan is the default express logger, just pulled out. :)

PR #50, opened 2016-05-13
Make PopulateVeteran more complete.

awong-dev:
Enters test data for nearly every field.

Data generated by filling out the form, then in the console with react-dev tools, selecting the <HealthCareApp> element and running:

```
> JSON.stringify($r.context.store.getState().veteran, null, 2)
```

And then running that through the following 2 regexps to make it lint cleanly:

```
  s/"\(\S*\)": /\1: /g
  s/"/'/g
```

### DIFF
--- a/src/client/reducers/veteran/index.js
+++ b/src/client/reducers/veteran/index.js
@@ -141,133 +141,443 @@ const blankVeteran = {
 
 export const completeVeteran = {
   veteranFullName: {
-    first: makeField('FirstName', true),
-    middle: makeField('MiddleName', true),
-    last: makeField('LastName', true),
-    suffix: makeField('Jr.', true),
+    first: {
+      value: 'FirstName',
+      dirty: true
+    },
+    middle: {
+      value: 'MiddleName',
+      dirty: true
+    },
+    last: {
+      value: 'LastName',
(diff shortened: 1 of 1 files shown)


src/client/reducers/veteran/index.js
```diff
@@ -141,133 +141,443 @@ const blankVeteran = {
...
+  email: {
+    value: 'foo@example.com',
+    dirty: false
+  },
+  emailConfirmation: {
+    value: 'foo@example.com',
+    dirty: false
+  },
+  homePhone: {
+    value: '123-124-1234',
```
crwallace: The validations for phone number have changed so that it no longer accepts dashes.

src/client/reducers/veteran/index.js
```diff
@@ -141,133 +141,443 @@ const blankVeteran = {
...
+    dirty: false
+  },
+  mobilePhone: {
+    value: '123-555-1234',
```
crwallace: same note here about dashes for phone validation.

crwallace:
Small nit but other once fixed LGTM!

awong-dev:
Fixed! committing.

PR #51, opened 2016-05-13
Match up loopback mount and api root paths.

awong-dev:
The Loopback server expects to be mounted on '/'.  The 'restApiRoot' is
then used to filter requests down.

Previously, the server was mounted under /hca which confused it because
the generated paths were expecting /v1/api/{resource} but express was
configured to route to /hca/v1/api{resource}.


robbiethegeek:
LGTM

PR #52, opened 2016-05-13
#220 Veteran's Personal Info validations including tests.

robbiethegeek:


src/server/utils/validations.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+const moment = require('moment');
+const _ = require('lodash');
+
```
awong-dev: spurious newline :)

src/server/utils/validations.js
```diff
@@ -0,0 +1,44 @@
...
+
+module.exports = {
```
awong-dev: Should we be using ES2015 module syntax now that we're in node 4.4.4?

test/server/enrollment-system.spec.js
```diff
@@ -0,0 +1,33 @@
+const chai = require('chai');
+chai.should();
```
awong-dev: I've been pushing to insert a newline between 3rd party includes and local ones.

src/server/utils/validations.js
```diff
@@ -0,0 +1,44 @@
...
+    if (_.isEmpty(data) || !_.isString(data)) {
+      return '';
+    }
+    const validatedString = _.upperFirst(data);
+    if (count && _.isNumber(count)) {
+      return (validatedString.length > count) ? validatedString.slice(0, count) : validatedString;
+    }
+    return validatedString;
+  },
+  validateSsn: (inputSsn) => {
```
awong-dev: Can we add some JSDoc comment that summarize the the format validation?  Reading a regexp is really hard.

test/server/enrollment-system.spec.js
```diff
@@ -0,0 +1,33 @@
...
+const enrollmentSystem = require('../../src/server/enrollment-system');
+
+describe('enrollment-system base tests', () => {
+  describe('validate the input is a non-empty Object', () => {
+    it('should return an empty object when nothing is passed in', () => {
+      const result = enrollmentSystem.veteranToSaveSubmitForm();
+      result.should.be.empty;
```
awong-dev: Also, I had pushed using expect() instead of the should DSL in the client code.  But I don't feel strongly here. Just an FYI.

test/server/utils/validations.spec.js
```diff
@@ -0,0 +1,123 @@
+const validations = require('../../../src/server/utils/validations');
+const chai = require('chai');
+chai.should();
+describe('validations', () => {
```
awong-dev: newline before `describe()` blocks?

test/server/utils/validations.spec.js
```diff
@@ -0,0 +1,123 @@
...
+      });
+      it('if the date passed in is a string after today', () => {
+        const convertedDate = validations.dateOfBirth('dec 1, 2016');
+        convertedDate.should.be.a('string');
+        convertedDate.should.be.empty;
+      });
+    });
+    describe('should return a validated string', () => {
+      it('if the date passed in is a string before today', () => {
+        const convertedDate = validations.dateOfBirth('dec 1, 1974');
```
awong-dev: I thought we always were expecting a broken apart object. Do we need to be parsing date strings?

awong-dev:
Overall looking solid.

I like the test pattern.  Main non-nitpicky question is about whether we should use ES2015 module syntax.

awong-dev:
My bias here would be to get it checked in soon (it looks already like it's near a good stopping point even if not complete) so the rest of us can pile on to the files.

test/server/utils/validations.spec.js
robbiethegeek: //  * personInfo / dob, mm/dd/yyyy; cannot be in the future, Yes, "Element is accepted but not used as this element will not be stored in ADR,  it must come from MPI (the authoritative source)."
so I started with the assumption that we can generate a string from the JSON coming in (even if that is `veteran.veteranDateOfBirth.month/veteran.veteranDateOfBirth.day/veteran.veteranDateOfBirth.year`

awong-dev: This is something to ask @alexose to confirm, but AFAIK, we are passing in dates as a JSON hash of the shape `{month, day, year}`.

src/server/utils/validations.js
```diff
@@ -0,0 +1,68 @@
...
+const _ = require('lodash');
+
+const validations = {
+  /**
+   * Validate a string as and convert to a valid format.
+   * @param {string} inputDob The date of birth string to validate / convert.
+   * @returns {string} If input was valid a valid SSN if not valid an empty string.
+   */
+  dateOfBirth: (inputDob) => {
+    let convertedDob = '';
```
awong-dev: I feel like you could elide this variable by just `return '';` in the error cases and returning the `moment()` call directly.

src/server/utils/validations.js
```diff
@@ -0,0 +1,68 @@
...
+      return '';
+    }
+    const validatedString = _.upperFirst(data);
+    if (count && _.isNumber(count)) {
+      return (validatedString.length > count) ? validatedString.slice(0, count) : validatedString;
+    }
+    return validatedString;
+  },
+  /**
+   * Validate a string or number to meet the requirements offrom the original
```
awong-dev: nit offfrom -> of

src/server/utils/validations.js
```diff
@@ -0,0 +1,68 @@
...
+   * 1010ez pdf form.
+   * Conditions for valid SSN :
```
awong-dev: Nice comment!

awong-dev:
LGTM w/caveats

I added a couple of comments, but given the churn-rate on this file, I'm going to bias on merging so we can reduce conflicts.  Please address in separate PR.

src/server/utils/validations.js
```diff
@@ -0,0 +1,68 @@
...
+      return convertedDob;
+    }
+    const parsedDob = new Date(inputDob);
+    if (moment().isBefore(parsedDob)) {
+      return convertedDob;
+    }
+    convertedDob = moment(parsedDob).format('MM/DD/YYYY');
+    return convertedDob;
+  },
+  /**
```
awong-dev: I was sorta expect function statement. For example:

```
function validateString() {
}

module.exports = { validateString, dateOfBirth, ...};
```

It's more verbose but feels a bit more explicit.  Push back if it annoys you. This kind of style is purely preference and I don't want to dictate...

test/server/enrollment-system.spec.js
robbiethegeek: don't think we can do this cause of the inability to use es2015 module import / export

PR #54, opened 2016-05-13
Reduce veteran JSON a little further

alexose:
Per @awong-dev 's request, this turns any field that looks like this:

`"veteranSocialSecurityNumber": {
    "value": "",
    "dirty": false
  },`

into

`"veteranSocialSecurityNumber": "",`


awong-dev:
LGTM

PR #55, opened 2016-05-14
Wire up form -> SOAP. End2End works now!

awong-dev:
Fix up some of the request format issues and normalize the test data so
it passes the SOAP validation rules.

With this CL, if you have the right endpoint uncommented in voa-rest.js,
then when on the VA network, it is possible to
  (1) load the UI with `devPanel=1`
  (2) click the "Populate Veteran" button
  (3) Navigate to the "review-and-submit" URL
  (4) Click "Submit Application" on the bottom of the form
  (5) Successuflly sumbit an application to the dev ES system.

There's still lots of validations to do, but at least the proof of
concept works now.

### DIFF
--- a/hca-api-stub/voa-rest.js
+++ b/hca-api-stub/voa-rest.js
@@ -75,6 +75,8 @@ function attach(app) {
       remotingEnabled: true,
       wsdl: path.join(__dirname, './voa.wsdl'),
       url: endpoint.esPreprod,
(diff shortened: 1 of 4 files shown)


alexose:
Yep, this looks good.  As we work on `src/server/enrollment-system.js`, let's keep an eye out for places to simplify... this is the sort of thing that has the tendency to get unreadable in a hurry.

👍

PR #56, opened 2016-05-16
Create new sections and edit content

akainic:
Create additional sections where an existing one was too long. Also changed some other content.

### DIFF
--- a/src/client/components/household-information/FinancialDisclosureSection.jsx
+++ b/src/client/components/household-information/FinancialDisclosureSection.jsx
@@ -33,17 +33,25 @@ class FinancialDisclosureSection extends React.Component {
       content = (<fieldset>
         <legend>Financial Disclosure</legend>
         <div className="input-section">
-          <p>In the next 4 sections you will be asked to provide your financial information from the most recent tax year. This
-          information is used to determine your eligibility for health care. Your most recent
-          tax filing will have all the information you need in it.</p>
+          <p>You will now be asked to provide your financial information from the
+          most recent tax year. We ask for this information for 2 reasons:</p>
+
+          <ol>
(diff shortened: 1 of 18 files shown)


crwallace:
Love the new demographic layout. I read an article this weekend that said that the next 2020 census will read, "Which categories best describe you" in an effort to not say "race" or "origin". I think it might make sense for us to mimic that here and ask in the same way.

akainic:
@crwallace Love that idea, I'll change the language we have for that question now

src/client/components/household-information/FinancialDisclosureSection.jsx
```diff
@@ -33,17 +33,24 @@ class FinancialDisclosureSection extends React.Component {
       content = (<fieldset>
         <legend>Financial Disclosure</legend>
         <div className="input-section">
-          <p>In the next 4 sections you will be asked to provide your financial information from the most recent tax year. This
-          information is used to determine your eligibility for health care. Your most recent
-          tax filing will have all the information you need in it.</p>
+          <p>You will now be asked to provide your financial information from the
+          most recent tax year. We ask for this information for 2 reasons:</p>
+
+          <ol>
```
crwallace: This looks great and makes perfect sense to me. Lets go ahead and link the words "qualifying eligibility factor" with the hover list.

akainic: Awesome. I'm going to add a TODO to tackle that next so we can get this PR merged in faster (I think that might take a bit of time, since I'll have to create a new tooltip component for that).

crwallace:
@akainic This looks great with a couple of add-ons.

PR #57, opened 2016-05-16
fixes routes on review panels.

crwallace:


akainic:
What are we doing with the panel ids?

akainic:
LGTM!

PR #59, opened 2016-05-16
Make additional updates to validations and conditional logic

akainic:
Additional fixes and the conditional logic we need.


crwallace:
LGTM!

PR #60, opened 2016-05-16
Adds Submit Message and Preloader

crwallace:


hca-api-stub/voa-rest.js
```diff
@@ -74,7 +74,7 @@ function attach(app) {
       connector: require('loopback-connector-soap'),
       remotingEnabled: true,
       wsdl: path.join(__dirname, './voa.wsdl'),
-      url: endpoint.esPreprod,
+      url: 'http://vaausesrapp803.aac.va.gov:7401/voa/voaSvc',
```
akainic: Is this a permanent change or just a temporary fix? Why not also do `endpoint.c7401`?

hca-api-stub/voa-rest.js
```diff
@@ -10,17 +10,17 @@ function attach(app) {
...
-    c7401: 'http://vaausesrapp803.aac.va.gov:7401/voa/voaSvc',
-    e6401: 'https://vaausesrapp803.aac.va.gov:6401/voa/voaSvc',
-    // ES "official" endpoints for development
-    esSqa: 'https://vaww.esrstage1a.aac.va.gov/voa/voaSvc',
-    esDev: 'https://vaww.esrdev30.aac.va.gov:8432/voa/voaSvc',
-    // ES endpoints that require keys from PKI team and certs installed by Joshua
-    esPreprod: 'https://vaww.esrpre-prod.aac.va.gov/voa/voaSvc',
-    esProd: 'https://vaww.esr.aac.va.gov/voa/voaSvc'
-  };
+  // const endpoint = {
```
akainic: Did you mean to leave these commented out?

public/index.html
```diff
@@ -107,7 +107,10 @@ <h2 class="sr-only">Apply for Health Care</h2>
 <div class="main" role="main">
   <div class="section">
     <div id="react-root">
-      Ruh-roh. React not loading.
+      <div class="loading-message">
+        <h3>Please wait while we load the application for you.</h3>
```
akainic: So happy with this :)

src/client/components/HealthCareApp.jsx
```diff
@@ -109,18 +107,30 @@ class HealthCareApp extends React.Component {
       if (!response.ok) {
         throw new Error(response.statusText);
       }
-      store.dispatch(updateSubmissionStatus('submitSucceeded', response.json()));
+      response.json().then(data => {
+        this.props.onUpdateSubmissionStatus('applicationSubmitted', data);
+        console.log('data', data);
```
akainic: Should we remove this `console.log`?

src/client/components/HealthCareApp.jsx
```diff
@@ -220,11 +266,44 @@ class HealthCareApp extends React.Component {
...
+function mapStateToProps(state) {
+  return {
+    data: state.veteran,
+    uiState: state.uiState,
+  };
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    // onStateChange: (field, update) => {
```
akainic: Can we remove this?

src/client/components/HealthCareApp.jsx
```diff
@@ -109,18 +107,30 @@ class HealthCareApp extends React.Component {
...
+      }, 8000);
     }).catch(error => {
-      store.dispatch(updateSubmissionStatus('submitFailed', error));
+      setTimeout(() => {
+        this.props.onUpdateSubmissionStatus('submitFailed', error);
+      }, 5000);
     });
 
-    this.scrollToTop();
+    // this.scrollToTop();
```
akainic: Can probably remove this

src/client/components/HealthCareApp.jsx
```diff
@@ -144,32 +154,68 @@ class HealthCareApp extends React.Component {
...
           <div className="small-6 medium-5 columns">
             {nextButton}
           </div>
         </div>
       );
+    } else if (this.props.location.pathname === '/submit-message') {
+      buttons = (
+        <div className="row progress-buttons">
+          <div className="small-6 medium-5 columns">
+            <button className="usa-button-primary">Back to Main Page</button>
```
akainic: Doesn't look like this button actually goes anywhere now

src/client/components/HealthCareApp.jsx
```diff
@@ -109,18 +107,30 @@ class HealthCareApp extends React.Component {
...
+      });
+      setTimeout(() => {
+        this.context.router.push(this.getUrl('next'));
+        this.scrollToTop();
+      }, 8000);
     }).catch(error => {
-      store.dispatch(updateSubmissionStatus('submitFailed', error));
+      setTimeout(() => {
+        this.props.onUpdateSubmissionStatus('submitFailed', error);
+      }, 5000);
```
akainic: I feel like something else should happen besides just changing the button text if the submission fails. Perhaps a message should appear below the buttons explaining that their submission failed and why it might have failed? At first I thought perhaps the submitMessage component should handle both cases and either way you should be redirected to that page, either with details of your success or something about how your application could not be submitted at this time. But I also see value of staying on the same page because otherwise they will lose all of their data. Not sure what the best solution is here, but some other messaging would probably help.

hca-api-stub/voa-rest.js
crwallace: when we were trying to get this to work @awong-dev told me to replace `url: endpoint.esPreprod` with `url: 'http://vaausesrapp803.aac.va.gov:7401/voa/voaSvc'` so I didn't even look at what was in the block. I can swap this out.

hca-api-stub/voa-rest.js
crwallace: Commented this out because it was giving me a lint error because we were not using `endpoint`. Changing that now though.

src/client/components/HealthCareApp.jsx
crwallace: Yep! Forgot to remove after debugging.

src/client/components/HealthCareApp.jsx
crwallace: No this was something that I was hoping @EmilyTav and @staceylanger would comment on in our walk-through. I wasn't sure what we wanted the user to do once they have completed and where it should take them. What does "Main Page" mean? Is that `vets.gov` or `vets.gov/health-care/`? Would love thoughts.

akainic: Makes sense! I have no strong opinions about this. Let's definitely check with @staceylanger since this is a navigation issue.

src/client/components/HealthCareApp.jsx
crwallace: Agreed! But right now we don't get anything useful back from the endpoint about why the submission didn't go through so I have nothing to show the user at this point. We could add some place holder text for now and just say "Please review your application and make sure that you have completed ALL required field before trying to resubmit."

src/client/components/HealthCareApp.jsx
```diff
@@ -84,17 +85,14 @@ class HealthCareApp extends React.Component {
...
     // Strip out unnecessary fields that track UI state
     function reducer(i, d) {
       return typeof d.value !== 'undefined' ? d.value : d;
     }
 
-    store.dispatch(updateSubmissionStatus('submitPending'));
-    store.dispatch(updateCompletedStatus(path));
+    this.props.onUpdateSubmissionStatus('submitPending');
 
     // POST data to endpoint
```
akainic: Let's first check if there are any validation errors in the form, and then only post data conditionally on that being false.

akainic:
@crwallace I was getting this same build error yesterday. When I rebase and squashed, and then forced push, it fixed it. Not sure why this is failing, but you can try that. Otherwise, LGTM!

robbiethegeek:
I restarted the build on Travis.

src/client/components/HealthCareApp.jsx
staceylanger: If either of you could get me a way to see this page I can weigh in. I haven't been able to successfully submit in staging.

akainic: @staceylanger That's because we're not actually able to submit yet from staging, sorry about that! @crwallace could you provide Stacey with a screenshot of the submit page?

awong-dev: Alternately, it turns out the devPanel is available on staging still.

@staceylanger use https://staging.vets.gov/healthcare/apply/?devPanel=1 which will add 3 (very ugly) widgets to the top of the page that will let you navigate to any part of the site ignoring all the checks and flows.

akainic: Oh awesome! Thanks @awong-dev!

PR #64, opened 2016-05-17
Incorrect Object typecheck broke submission.

awong-dev:
typeof veteran !== Object is incorrect because typeof veteran returns a
string.  typeof veteran !== 'object' is probably correct, but might as
well stay in lodash land since it's already in this expression.


alexose:
LGTM.

robbiethegeek:
Thanks for catching this @awong-dev!

PR #65, opened 2016-05-17
tiny fix + comments for applicationInfo

alexose:


alexose:
I put this in a separate PR because I wasn't sure about the implications of using the time on the server to determine `appDate`.

Is there a chance that it'll cause problems if there's an off-by-one error due to timezones?

alexose:
Also, I noticed that we pulled in moment.js on the validation side of things.  So we can probably get rid of https://github.com/department-of-veterans-affairs/healthcare-application/blob/master/src/server/enrollment-system.js#L83-L96

robbiethegeek:
LGTM

robbiethegeek:
is there someone that can answer your Question @alexose  that you are thinking of regarding the timezones?

awong-dev:
@alexose I believe this is an audit timestamp and that we really have no sane choice other than to use server timestamp.  What we can ensure in moment is that we're always stamping with a consistent timezone that doesn't have daylight savings as opposed to whatever the server is configured with.

The right thing to do is to ping Joshua.

awong-dev:
@alexose Let's merge this in. I will create bug for following up about timezones.

robbiethegeek:
bump, where does this live in the newly refactored API?

awong-dev:
@robbiethegeek same place.

robbiethegeek:
Just updated branch can you squash and then will merge in if all is good.

PR #66, opened 2016-05-18
Language edits

akainic:
Most of the content changes we discussed with @staceylanger, @EmilyTav, and @crwallace.

Main changes:
1. Typo fixes
2. General language changes
3. Changing many checkboxes into radio buttons
4. Adding alerts for financial disclosure page
5. Removing hint for DateInput
6. Fixing problem with VA Medical Facility not displaying if you go back to that section
7. Fixing conditional logic in Spouse section
8. Update validations accordingly

### DIFF
--- a/src/client/components/household-information/DeductibleExpensesSection.jsx
+++ b/src/client/components/household-information/DeductibleExpensesSection.jsx
@@ -17,19 +17,7 @@ function getErrorMessage(field, message) {
 class DeductibleExpensesSection extends React.Component {
   render() {
     const message = 'Please enter only numbers and a decimal point if necessary (no commas or currency signs)';
-    let notRequiredMessage;
     let content;
-
-    if (this.props.data.receivesVaPension === true) {
-      notRequiredMessage = (
-        <p>
-          <strong>
-            You are not required to enter financial information because you
-            indicated you are receiving a VA pension.
-          </strong>
-        </p>
-      );
-    }
 
     if (this.props.isSectionComplete && this.props.reviewSection) {
       content = (<table className="review usa-table-borderless">
@@ -55,8 +43,6 @@ class DeductibleExpensesSection extends React.Component {
     } else {
       content = (<fieldset>
         <legend>Previous calendar year deductible expenses</legend>
-        {notRequiredMessage}
-
         <p>
           Tell us a bit about your expenses this past calendar year. Enter information for any expenses that apply to you.
         </p>
--- a/src/client/components/va-benefits/VAInformationSection.jsx
+++ b/src/client/components/va-benefits/VAInformationSection.jsx
@@ -35,31 +35,22 @@ class VAInformationSection extends React.Component {
         );
     } else {
       content = (<fieldset>
-        {/* TODO: Change the headings to something related to the questions. */}
-        <legend>Veteran</legend>
-        <p>
-          Please review the following list and select all the responses that apply to you.
-          This information will be used to determine which sections of the Application for
-          Health Benefits you should complete.
-        </p>
+        <legend>VA Benefits</legend>
         <p>(<span className="hca-required-span">*</span>) Indicates a required field</p>
         <div className="input-section">
           <ErrorableRadioButtons required
-              errorMessage={validateIfDirty(this.props.data.isVaServiceConnected, isNotBlank) ? '' : 'Please select a response'}
-              label="Are you VA Service Connected 50% to 100% Disabled?"
-              options={yesNo}
-              value={this.props.data.isVaServiceConnected}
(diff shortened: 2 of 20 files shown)


src/client/components/household-information/DeductibleExpensesSection.jsx
```diff
@@ -56,7 +56,6 @@ class DeductibleExpensesSection extends React.Component {
       content = (<fieldset>
         <legend>Previous calendar year deductible expenses</legend>
         {notRequiredMessage}
```
crwallace: Are we still passing a `notRequiredMessage` to the user?

src/client/components/va-benefits/VAInformationSection.jsx
```diff
@@ -35,31 +35,22 @@ class VAInformationSection extends React.Component {
...
               options={yesNo}
               value={this.props.data.compensableVaServiceConnected}
               onValueChange={(update) => {this.props.onStateChange('compensableVaServiceConnected', update);}}/>
-          <span>
-            A VA determination that a Service-connected disability is severe enough to warrant monetary compensation.
-          </span>
+
+          <ErrorableRadioButtons required
+              errorMessage={validateIfDirty(this.props.data.isVaServiceConnected, isNotBlank) ? '' : 'Please select a response'}
+              label="Are you currently receiving monetary compensation for a service-connected disability with a rating of 50% to 100% disabled?"
```
crwallace: Did this look bad when you tried 
```
Are you currently receiving monetary compensation for a service-connected disability with a rating of:
0-40
50-100
```

src/client/components/va-benefits/VAInformationSection.jsx
```diff
@@ -35,31 +35,22 @@ class VAInformationSection extends React.Component {
...
+              label="Are you currently receiving monetary compensation for a service-connected disability with a rating of 0% - 40% disabled?"
               options={yesNo}
               value={this.props.data.compensableVaServiceConnected}
               onValueChange={(update) => {this.props.onStateChange('compensableVaServiceConnected', update);}}/>
-          <span>
-            A VA determination that a Service-connected disability is severe enough to warrant monetary compensation.
-          </span>
+
+          <ErrorableRadioButtons required
+              errorMessage={validateIfDirty(this.props.data.isVaServiceConnected, isNotBlank) ? '' : 'Please select a response'}
```
crwallace: Is there a gap here? What happened to 41% to 49%?

crwallace:
LGTM with a couple of questions.

src/client/components/va-benefits/VAInformationSection.jsx
akainic: Yeah, it's not great. The reason I kept it this way is because I was encountering some problems doing it the ways we talked about yesterday: 
1. I couldn't bold just the percentages because the whole label is being passed as a prop on ErrorableRadioButtons, so I didn't know how to bold just part of it. 
2. I was worried about accessibility issues if I extracted the main content of the label and had the labels simply be the percentages. Would that make sense to someone using a screen reader? I'm guessing not. 

If you have suggestions for how to handle this differently I'm all ears! Unfortunately this was the best I could come up with for now.

src/client/components/household-information/DeductibleExpensesSection.jsx
akainic: Good point, what we're passing is kind of messed up. I'm just going to remove these notRequiredMessages for now because I think we're taking care of this in the FinancialDisclosure language.

src/client/components/va-benefits/VAInformationSection.jsx
crwallace: That makes sense. It seemed weird to me but to someone who know what their rating is this wouldn't be weird. @maryannbrody do you have any thoughts on the literacy question.

src/client/components/va-benefits/VAInformationSection.jsx
crwallace: To solve for accessibility issues you could do something like:
```
<fieldset>
<legend>Are you currently receiving monetary compensation for a service-connected disability with a rating of:</legend>
< ErrorableRadioButtons label ="0-40" ... />
< ErrorableRadioButtons label ="50-100" ... />
</fieldset>
```

I don't think this is super crucial so leaving it like it is, is also a fine option as well.

src/client/components/va-benefits/VAInformationSection.jsx
maryannbrody: Hello @crwallace & @akainic -- here is what i recommend:

- Do you currently receive monetary compensation (pay) from the VA for a service-connected disability with a rating of 10, 20, 30, or 40 percent?
- Do you currently receive monetary compensation (pay) from the VA for a service-connected disability with a rating of 50 percent or more?

PR #69, opened 2016-05-18
Set dev endpoint as the default.

awong-dev:
Most of our testing should be against dev anyways. Might as well make
this the default.


robbiethegeek:
Just updated branch via Github, if the tests pass I will LGTM and merge

PR #70, opened 2016-05-19
Migrate most of mapper.js into enrollment-system.js

awong-dev:
Walked through all of blankVeteran in src/common/veteran.js and put in a
basic mapping inside enrollment-system.js for the field. Most of the
code was lifted from Joe Stewart's src/client/util/mapper.js.

All unmigrated fields have been marked with a comment in
src/common/veteran.js citing the work that needs to be done.

This is sorta messy. Most of the new utility functions should be moved
into src/server/util/validation.js. The next step is to make that move
and add unittests.


robbiethegeek:
LGTM from me and @awong-dev to get this in and then will branch and allow us to parallelize fixes.

PR #71, opened 2016-05-20
Redirect to landing if previous sections are not complete

akainic:
There are 2 scenarios where we want to redirect users to the introduction page (to prevent them from breaking the form logic that depends on moving linearly through the form): 
1. Force refresh
2. Going directly to a url

Not sure the way I'm handling this for the second scenario is ideal.

### DIFF
--- a/src/client/components/HealthCareApp.jsx
+++ b/src/client/components/HealthCareApp.jsx
@@ -28,6 +28,15 @@ class HealthCareApp extends React.Component {
     this.handleContinue = this.handleContinue.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.getUrl = this.getUrl.bind(this);
+    this.redirect = this.redirect.bind(this);
+  }
+
+  componentWillMount() {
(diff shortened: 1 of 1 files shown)


src/client/components/HealthCareApp.jsx
```diff
@@ -65,6 +74,26 @@ class HealthCareApp extends React.Component {
     });
   }
 
+  redirect() {
+    const previousPaths = [];
+    const currentPath = this.props.location.pathname;
+    const paths = _.keys(this.props.uiState.sections);
+
+    for (let i = 0; i < paths.length; i++) {
```
robbiethegeek: I think this is something that could be written a bit more effectively with a map but if this is working I approve, don't want to hold this up by prematurely optimizing.

robbiethegeek:
LGTM

src/client/components/HealthCareApp.jsx
akainic: Added a TODO!

PR #72, opened 2016-05-20
Map a few more fields.

awong-dev:
Marriage "start date" huh? Sounds like a job.


awong-dev:
I've verified that submit still works.

robbiethegeek:
LGTM

PR #73, opened 2016-05-20
Merge childrenIncome into children in React store

awong-dev:
This makes it much easier to manipulate the model since there are not
any parellel arrays.

Parallel arrays === bad.

PR #75, opened 2016-05-20
Dependent info incomes

awong-dev:
Implement dependentInfo incomes field.
Represents all the child incomes.


awong-dev:
@robbiethegeek This depends on #72 and #73 as it's rebased on top of a hypothetical merge of the two.

awong-dev:
I've verified that submit still works.

robbiethegeek:
LGTM

PR #76, opened 2016-05-20
Use browserHistory instead of hashHistory.

awong-dev:
fixes
https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/207.

### DIFF
--- a/src/client/react-entry.jsx
+++ b/src/client/react-entry.jsx
@@ -1,19 +1,25 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { IndexRedirect, Route, Router, hashHistory } from 'react-router';
+import { createHistory } from 'history';
+import { IndexRedirect, Route, Router, useRouterHistory } from 'react-router';
 import { Provider } from 'react-redux';
 import { createStore } from 'redux';
 
 import HealthCareApp from './components/HealthCareApp.jsx';
+import config from '../../config';
 import reducer from './reducers';
 import routes from './routes.jsx';
 
 const store = createStore(reducer);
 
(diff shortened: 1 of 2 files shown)


awong-dev:
@robbiethegeek for the node.js express change.

robbiethegeek:
Express change looks fine. Without tests please just test manually to confirm. Once that is confirmed LGTM on server.js

akainic:
Looks good except that the redirect to `/introduction` in HealthCareApp isn't working. If I go to just localhost:3000/healthcare/apply I get a 404.

awong-dev:
@akainic Nice catch!

I fixed.

awong-dev:
Rebased and fixed router ordering problems so now this should all work.

awong-dev:
Reassigning to @akainic to verify that my fix actually worked for her. If it works, let's merge.

akainic:
~~Works for me now!~~ Not yet, was too excited

awong-dev:
Okay, the problem is `npm run watch` is broken cause the hot reload path has a different server configruation. See the `historyApiFallback` option to the dev server...I need to figure out how to weave that through right.

awong-dev:
Apparently the solution is `historyApiFallback: true` to make webpack dev server do the equiv of serving the index page for every 404.

Yay?

Okay, this should work now.  @akainic, can you test in both `npm run watch` mode as well as `npm run webpack-noopt; npm run start` mode?  Make sure to do a `killall npm; killall node` to be sure...also might delete everything in `generated` before running `npm run webpack-noopt`.

akainic:
@awong-dev It worked for me with both `npm run watch` and `npm run webpack-noopt; npm run start`!

PR #77, opened 2016-05-20
Extract server config into a config file.

awong-dev:
Makes it easier to set new ports, certs, etc.

connected to: https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/256


alexose:
No objections.  If we do end up needing to keep sensitive information in here (passwords, etc), let's go down the config.js.example route.

👍

PR #78, opened 2016-05-20
Complete the mappings.

awong-dev:
With this, the mappings should be done.  Next is to verify with ES and
do bug fixing + tests.

### DIFF
--- a/src/common/veteran.js
+++ b/src/common/veteran.js
@@ -67,13 +67,13 @@ const blankVeteran = {
     year: makeField(''),
   },
   dateOfMarriage: {
-    month: makeField(''),  // TODO(awong): Ignored by ES System
-    day: makeField(''),  // TODO(awong): Ignored by ES System
-    year: makeField('')  // TODO(awong): Ignored by ES System
+    month: makeField(''),
+    day: makeField(''),
+    year: makeField('')
   },
   sameAddress: makeField(''),  // TODO(awong): Not sure how to handle the mapping.
-  cohabitedLastYear: makeField(''),  // TODO(awong): This should be scoped to spouse.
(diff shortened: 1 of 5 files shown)


src/common/veteran.js
```diff
@@ -67,13 +67,13 @@ const blankVeteran = {
...
-    day: makeField(''),  // TODO(awong): Ignored by ES System
-    year: makeField('')  // TODO(awong): Ignored by ES System
+    month: makeField(''),
+    day: makeField(''),
+    year: makeField('')
   },
   sameAddress: makeField(''),  // TODO(awong): Not sure how to handle the mapping.
-  cohabitedLastYear: makeField(''),  // TODO(awong): This should be scoped to spouse.
-  provideSupportLastYear: makeField(''),  // TODO(awong): This should be scoped to spouse.
+  cohabitedLastYear: makeField(''),  // TODO(awong): This name should be scoped to spouse.
```
robbiethegeek: What is still left on this set of TODOs?

robbiethegeek:
One minor question that is about a TODO so I am LGTM and merging this in.  @awong-dev take a look at the question about the TODO and if not captured as an issue add to the backlog?

src/common/veteran.js
awong-dev: Mostly I want to shift it into a `spouse {}` object.  But I think with the next PR, this should move into the API layer.

awong-dev: I consider this part of validation + testing as I think unittesting will make us really want to refactor the object structure a little.

robbiethegeek: makes sense was just trying to grok the todo :) thanks!

PR #79, opened 2016-05-21
Rip out Loopback. Use json schema for API.

awong-dev:
This removes all off loopback in favor of directly coding the rest
endpoint in express.

The impetus was loopback-connector-soap abstracted away the ability to
handle the certificate authority making the TLS setup harder. Also, the
connector's function was to bridge the internal SOAP API back outwards
which is functionality we don't want.

The new code is more straight forward.

Furthermore, the REST API has been changed so that submissions happen via
a POST to

  /api/hca/v1/applications

Status checks will eventually be exposed via a GET on

  /api/hca/v1/application/[:id]

which makes more sense based on REST semantics.

The applications schema is now defined in src/common/schema/application.js
using jsonschema. This gives a harder split between frontend model and
rest model allowing for the REST model to have stricter type checking that
makes the mapping code easier.

Code was tested against the characterization test. Actual unittests should be
added to replace the black-box characterization test.

As of right now, pre-prod connectivity is slighty broken because the
wsdl download fails the strict SSL checks. Debugging the CA problems is
the next step.


src/common/veteran.js
```diff
@@ -642,4 +643,68 @@ const completeVeteran = {
...
+      case 'provideSupportLastYear':
+      case 'receivesVaPension':
+      case 'provideFinancialInfo':
+      case 'understandsFinancialDisclosure':
+      case 'sameAddress':
+      case 'cohabitedLastYear':
+      case 'isCoveredByHealthInsurance':
+      case 'isMedicaidEligible':
+      case 'isEnrolledMedicarePartA':
+        return value.value === 'Y';
```
robbiethegeek: by default are the values N if not submitted or will these just become undefined?

src/common/veteran.js
```diff
@@ -642,4 +643,68 @@ const completeVeteran = {
...
+      case 'grossIncome':
+      case 'netIncome':
+      case 'otherIncome':
+      case 'deductibleEducationExpenses':
+      case 'deductibleFuneralExpenses':
+      case 'deductibleMedicalExpenses':
+        return Number(value.value);
+
+      default:
+        // fall through.
```
robbiethegeek: may trigger an eslint error

src/common/veteran.js
```diff
@@ -642,4 +643,68 @@ const completeVeteran = {
...
+    }
+
+    // Turn date fields into ISO8601 dates.
+    if (value.day !== undefined && value.month !== undefined && value.year !== undefined) {
```
robbiethegeek: we should use moment if it is still in the file or readd : http://momentjs.com/docs/#/displaying/as-iso-string/

src/server.js
```diff
@@ -41,17 +41,17 @@ function makeServer() {
...
+  return app;
 }
 
-const server = makeServer();
-const api = loopback();
-voaRest.attach(api);
-server.use('/', api);
+const app = makeApp();
+app.use(bodyParser.json());
+app.post(`${config.apiRoot}/applications`, api.submitApplication);
```
robbiethegeek: why both routes to the same method?

src/server/api.js
```diff
@@ -0,0 +1,72 @@
...
+  config.soap.wsdl,
+  {
+    security: securityArtifacts,
+    wsdl_options: securityArtifacts === null ? null : { // eslint-disable-line
+      rejectUnauthorized: false,
+      strictSSL: false,
+      requestCert: true
+    }
+  },
+  (err, client) => { voaService = client; });
```
robbiethegeek: we should deal with the error if we try to spin up and the server is offline it will blow up

src/server/api.js
```diff
@@ -0,0 +1,72 @@
...
+// TODO(awong): Remove config.url.
+
+function submitApplication(req, res) {
+  const contentType = req.get('Content-Type');
+  if (contentType !== 'application/json') {
+    res.status(400).json({ error: `Expects application/json content-type. Got "${contentType}"` });
+    return;
+  }
+
+  console.log(JSON.stringify(req.body, null, 2));
```
robbiethegeek: add todo this via a logger.

src/server/enrollment-system.js
```diff
@@ -263,12 +98,12 @@ function spanishHispanicToSDSCode(isSpanishHispanicLatino) {
...
-  if (veteran.isAsian) races.push({ race: '2028-9' });
-  if (veteran.isBlackOrAfricanAmerican) races.push({ race: '2054-5' });
-  if (veteran.isNativeHawaiianOrOtherPacificIslander) races.push({ race: '2076-8' });
-  if (veteran.isWhite) races.push({ race: '2106-3' });
-  return races;
+  if (veteran.isAmericanIndianOrAlaskanNative) races.push('1002-5');
+  if (veteran.isAsian) races.push('2028-9');
+  if (veteran.isBlackOrAfricanAmerican) races.push('2054-5');
+  if (veteran.isNativeHawaiianOrOtherPacificIslander) races.push('2076-8');
+  if (veteran.isWhite) races.push('2106-3');
```
robbiethegeek: is this the valid JSON for sequential elements?  Not sure of the format

robbiethegeek:
This looks good I don't have any real practical JSON schema experience, I am going to pass to @alexose for another set of eyes.

src/common/veteran.js
awong-dev: Tested locally. It doesn't. \o/

src/common/veteran.js
awong-dev: I will add a TODO but I had problems with timezones (as described on slack) such that I didn't want to transition through the native Date object.

src/common/schema/application.js
```diff
@@ -0,0 +1,293 @@
...
+      type: 'boolean'
+    },
+    radiumTreatments: {
+      type: 'boolean'
+    },
+    campLejeune: {
+      type: 'boolean'
+    }
+  }
+};
```
alexose: Without verifying that this is 100% schematically correct, I will say that this is awesome a big improvement 👍

src/common/veteran.js
alexose: Maybe a separate PR to transition these things to moment?

src/server.js
awong-dev: oops. that was a copy-pasta mistake.  The singular version is removed.

src/server/api.js
awong-dev: Yeah...I don't know how to do in node.js. Like what's the format of the `err` parameter? How do I reconnect? etc. etc.

Adding TODO now...

src/server/api.js
awong-dev: I'll just delete.

src/server.js
alexose: Backwards compatibility with loopback's shady plural conventions, I assume

src/server/enrollment-system.js
awong-dev: yup.

I couldn't find a spec or anything for the format, but it generates sensible looking xml when I capture it.

src/server/api.js
```diff
@@ -0,0 +1,72 @@
+'use strict'; // eslint-disable-line
+
+const fs = require('fs');
+const soap = require('soap');
+const debug = require('debug')('hca:api');
+const ajv = require('ajv');
```
alexose: `ajv` looks cool.  Approve.

alexose:
This looks great, @awong-dev! 👍

src/common/veteran.js
awong-dev: They will just become undefined. Which I think defaulting to false is okay?

src/common/veteran.js
awong-dev: I remain unconvinced that moment is a good idea for this particular action.

PR #81, opened 2016-05-23
Quick update to the veteran common file to send a valid required cohabitedLastYear.

robbiethegeek:


joesoftware:
LGTM after travis passes.

PR #83, opened 2016-05-23
Adding CLI tools for submit and status.

joesoftware:
Two simple tools added here for local back-end testing:

```
# Usage: ./submit.sh FILE
# Posts FILE (containing JSON data in the format used by the veteran form store)
# to a local instance of the backend server.
# Prints backend results upon completion.
```

```
# Usage: ./status.sh FORMSUBMISSIONID
# Creates a GET request from the provided FORMSUBMISSIONID.
# Prints backend results upon completion.
```


awong-dev:
LGTM.

The urls will have to change once my PR goes in, but not a big deal.

PR #85, opened 2016-05-24
Add test locking down the golden test data.

awong-dev:
This ensures the test data used in the UI is the same as the test data
file used for the API.  It does this by taking each test data blob and
ensuring a test exists that runs it through each validation/conversion
function to check that the exact same result is produced at each stage.

Also completes the JSON schema for children and insurance providers.


robbiethegeek:
LGTM

PR #86, opened 2016-05-24
Add basic status retrieval method.

joesoftware:


src/server/api.js
```diff
@@ -72,4 +72,25 @@ function submitApplication(req, res) {
...
+    res.status(500).json({ error: 'need id' });
+  }
+  const request = {
+    formSubmissionId: id
+  };
+  voaService.getFormSubmissionStatus(request, (err, response) => {
+    if (err) {
+      debug(`voaService response had error ${err}`);
+      // TODO(awong): This may leak server config info on error. Is that a problem?
+      res.status(500).send({ error: err });
```
awong-dev: aaand...I think I screwed this up slightly. This should be `.json({ error: err })` and not `.send()`.

The test is to look and see if the response header has `Content-Type: 'application/json'` appropriately set using `.send()`. If not, can we switch all these to .json?

robbiethegeek: pretty sure that .json is a syntactic sugar for forcing something non JSON to json, send accepts all formats and pivots the headers based on the response.
This is my general understanding

src/server.js
```diff
@@ -51,6 +51,7 @@ function makeApp() {
 
 const app = makeApp();
 app.use(bodyParser.json());
+app.get(`${config.apiRoot}/applications/:id`, api.getApplicationStatus);
```
robbiethegeek: Please take a look over at the PR I submitted about how to move the .get into the api.js where an Express router object is being returned  in general I like to create a single js file per API route (ie /application would be a single file and then /applications would be a 2nd js file).

src/server/api.js
joesoftware: I'll change all to `.json`.

src/server.js
joesoftware: Sounds good. I've left both in `api.js` since they're both attached to `/application`.

src/server/api.js
```diff
@@ -63,16 +63,42 @@ function submitApplication(req, res) {
       if (err) {
         debug(`voaService response had error ${err}`);
         // TODO(awong): This may leak server config info on error. Is that a problem?
-        res.status(500).send({ error: err });
+        res.status(500).json({ error: err });
```
awong-dev: Can you ping Robbie to ensure .json is right versus .send?  I *think* it's better to coerce to json in the event we make a programming error, but not sure.

awong-dev:
LGTM w/ one comment. Merging though. Fix in a follow-up PR if necessary.

src/server/api.js
joesoftware: Will do.

joesoftware: I tinkered with changing between `.json()` and `.send()` using `curl` as a client and the Chrome Boomerang app, and there was no observable difference. Both returned identical strings from JSON objects and both had _Content-Type: application/json_ in the returned headers.

awong-dev: Yup...I think Robbie was saying node in send() attempts to sniff the
content type.

What I'm wondering is if it's better to disable the sniffing by explicitly
using json().

PR #88, opened 2016-05-24
Adds fix for empty optional fields.

crwallace:


src/client/components/HealthCareApp.jsx
```diff
@@ -29,15 +29,15 @@ class HealthCareApp extends React.Component {
     this.handleContinue = this.handleContinue.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.getUrl = this.getUrl.bind(this);
-    this.redirect = this.redirect.bind(this);
+    // this.redirect = this.redirect.bind(this);
```
awong-dev: delete?

src/client/components/HealthCareApp.jsx
```diff
@@ -29,15 +29,15 @@ class HealthCareApp extends React.Component {
...
   }
 
   componentWillMount() {
-    this.redirect();
+    // this.redirect();
```
awong-dev: delete?

src/client/components/HealthCareApp.jsx
```diff
@@ -29,15 +29,15 @@ class HealthCareApp extends React.Component {
...
   }
 
   componentDidUpdate() {
-    window.addEventListener('hashchange', this.redirect());
+    // window.addEventListener('hashchange', this.redirect());
```
awong-dev: delete?

src/client/components/HealthCareApp.jsx
```diff
@@ -75,26 +75,26 @@ class HealthCareApp extends React.Component {
...
-    }
-
-    for (let j = 0; j < previousPaths.length; j++) {
-      if (this.props.uiState.sections[previousPaths[j]].complete === false) {
-        this.context.router.push('/introduction');
-        break;
-      }
-    }
-  }
+  // redirect() {
```
awong-dev: Just delete instead of commenting out.

src/common/schema/application.js
```diff
@@ -54,10 +54,21 @@ module.exports = {
         'zipcode',
       ]
     },
-    date: {
+    requiredDate: {
       format: 'date',
       type: 'string'
     },
+    optionalDate: {
```
awong-dev: Is this needed anymore if we know to remove the field at the client side before it's sent to the server?

src/common/schema/application.js
```diff
@@ -71,7 +82,14 @@ module.exports = {
           type: 'string'
         },
         suffix: {
-          'enum': options.suffixes
+          oneOf: [
```
awong-dev: Same here...can we just do this in the client side?

src/client/components/HealthCareApp.jsx
crwallace: @akainic asked me to just comment it out for now.

src/common/schema/application.js
```diff
@@ -167,7 +177,7 @@ module.exports = {
     },
     email: {
       type: 'string',
-      format: 'email'
+      pattern: '^(([a-zA-Z]|[0-9])|([-]|[_]|[.]))+[@](([a-zA-Z0-9])|([-])){2,63}[.](([a-zA-Z0-9]){2,63})+$|^$'
```
awong-dev: Can we document where this regexp comes from in a comment?

src/common/veteran.js
```diff
@@ -675,6 +675,22 @@ function veteranToApplication(veteran) {
...
+      // Optional Suffix fields
+      case 'veteranFullName':
+        if (value.suffix.value === '') { return undefined; }
+        break;
+
+      // Optional Date fields
+      case 'spouseDateOfBirth':
+      case 'dateOfMarriage':
+      case 'medicarePartAEffectiveDate':
+        if (value.day.value === '') { return undefined; }
```
awong-dev: Should this also test month/year to ensure they aren't empty?

awong-dev: Style nit: don't put all of the if statement on one line?

src/common/veteran.js
```diff
@@ -675,6 +675,22 @@ function veteranToApplication(veteran) {
...
+        break;
+
+      // Optional SSN fields
```
awong-dev: You can combine the Optional Suffix fields and Optional SSN fields.

The grouping here makes the most sense to be if they are done by "type" instead of semantic value.  Basically, we have things that look like dates, and things that look like strings.

src/common/veteran.js
```diff
@@ -675,6 +675,29 @@ function veteranToApplication(veteran) {
...
+      // Optional String fields
+      case 'spouseSocialSecurityNumber':
+        if (value.value === '') {
+          return undefined;
+        }
+        break;
+
+      case 'veteranFullName':
+      case 'spouseFullName':
+        if (value.suffix.value === '') {
```
akainic: Why only checking for `suffix` being blank and not also `middleName`?

src/common/schema/application.js
```diff
@@ -128,9 +128,19 @@ module.exports = {
...
+        }, {
+          type: 'string',
+          pattern: '^$'
+        }
+      ]
+    },
     phone: {
       type: 'string',
-      pattern: '^[0-9]{10}$'
+      pattern: '^[0-9]{10}$|^$'
```
markolson: Do you want to the phone number type itself is always allowed to be blank, or do we care more about specific uses of it (i.e. homePhone must exist, but spousePhone can be blank)?

crwallace: None of the phone numbers are required fields so no real need for that.

awong-dev: The pattern at the API layer should not accept blank.

The changes to `phone` and `state` should likely be reverted in favor of removing them from the object before it is even sent in the server.

awong-dev:
LGTM. Let's get this in...once I fix CI.... grrrrrrr

PR #89, opened 2016-05-24
Adding logging to the API calls that are in place.

robbiethegeek:


src/server/api.js
```diff
@@ -11,94 +11,108 @@ const validate = ajv({ allErrors: true, errorDataPath: 'property', removeAdditio
...
+      fs.accessSync(config.soap.clientCertPath, fs.R_OK);
+      fs.accessSync(config.soap.clientKeyPath, fs.R_OK);
+      return {
+        scheme: 'ClientSSL',
+        certPath: config.soap.clientCertPath,
+        keyPath: config.soap.clientKeyPath
+      };
+    } catch (ex) {
+      debug("Can't read security artifacts. Starting without certificate and key.");
+      options.logger.info("Can't read security artifacts. Starting without certificate and key.");
```
awong-dev: My main concern with the logger here is that we have some sort of easy way to parse the format. I assume that winston gives us a common header?

awong-dev:
LGTM with the one comment.

PR #90, opened 2016-05-24
Saucelabs integration

alexose:
This pull request integrates Travis + Nightwatch + Saucelab and cleans up some of webpack's configuration options.

There's a bit more to do (generating the production assets could be a little better.  Also I'm not 100% sure if this breaks people's dev environments or not), but this is most of it.


.travis.yml
```diff
@@ -1,37 +1,44 @@
 sudo: required
-
 language: node_js
 node_js:
-  - "4.4.4"
+- 4.4.4
```
awong-dev: why did we reindent the yml?  makes diffing a little annoying.

.travis.yml
```diff
@@ -1,37 +1,44 @@
...
-      - ubuntu-toolchain-r-test
+    - llvm-toolchain-precise-3.6
+    - ubuntu-toolchain-r-test
     packages:
-      - clang-3.6
-      - g++-4.8
+    - clang-3.6
+    - g++-4.8
+  sauce_connect:
+    username: alexose
```
awong-dev: do we have a dsva sauce labs account?

scripts/e2e-tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# For running integration tests in travis.  This should happen any time we merge to staging.
+npm run webpack-prod;
```
awong-dev: remove extra semicolons

scripts/e2e-tests.sh
```diff
@@ -0,0 +1,6 @@
...
+npm run serve &
+sleep 3;
```
awong-dev: eek. what about writing a node.js script that loads the server.js as a module and watches for a start signal or something?

As is, this is just asking for flaky runs on travis over time.

awong-dev:
Nifty. My main concern is with the sleep 3.  Would like to get that into something w/o the race condition before submission if it's not too hard.

scripts/e2e-tests.sh
robbiethegeek: Yeah I would have asked the reason for  the sleep..echoing Albert here :)

.travis.yml
alexose: Blah.  That happened when I used `travis encrypt --yes`.  I'll set it back to how it was

.travis.yml
alexose: Still waiting on that via @plusjeff

alexose:
GitHub appears to be comparing this branch against an older version of `master`.  I'm not sure why, but I created a new PR, and it seems to have fixed it:

https://github.com/department-of-veterans-affairs/healthcare-application/pull/92

PR #91, opened 2016-05-25
Correctly configure server certificate authority.

awong-dev:
With this, NODE_TLS_REJECT_UNAUTHORIZED=0 should no longer be required
to establish an encrypted connection with SOAP endpoints using CAs that
are "not generally trusted" (which includes all Federal CA and agency
internal CAs).

This is important for mitigating MiTM attacks on establishing the TLS
connection.


joesoftware:
LGTM. (Inspected and tested locally with my copy of the private key.)

robbiethegeek:
@awong-dev can you update the README.md to describe the requirements for local key and crt files.  If we need to manually create can you describe where they come from.

You are close to it and I feel like I can't wrap my head around what is still required or not.

awong-dev:
Back at ya Robbie

README.md
```diff
@@ -58,227 +58,135 @@ npm install
...
+staging and production servers. During normal development, test against `dev`
+or `sqa`.
+
+##### Selecting an ES endpoint
+
+The server endpoints are configured in `config.json` via the
+`config.soap.endpoint` and `config.soap.wsdl` keys.  See the `endpoint` object
+for a list of the common server endpoints to use.
+
+The configuration default s https SOAP request sent to `esDev`. By default,
```
markolson: "default is" ?

awong-dev: fixed.

robbiethegeek:
Thanks @awong-dev LGTM will merge when the tests pass.

PR #92, opened 2016-05-25
Saucelabs integration cleanup

alexose:
This is a continuation of #90, which I made because GitHub was being mean.


alexose:
Whoever merges this would be wise to squash...

scripts/e2e-tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# For running integration tests in travis.  This should happen any time we merge to staging.
+
+# Run webpack-prod and start server
+npm start &
+
+# Wait for server to begin accepting connections
+# via http://unix.stackexchange.com/questions/5277
+while ! echo exit | nc localhost 3000; do sleep 1; done
```
awong-dev: hah. cute.

awong-dev:
LGTM

PR #93, opened 2016-05-25
adding babel-polyfill to the webpack config.

robbiethegeek:


webpack.config.js
```diff
@@ -5,9 +5,10 @@ var webpack = require('webpack');
 var bourbon = require('bourbon').includePaths;
 var neat = require('bourbon-neat').includePaths;
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
+require('babel-polyfill');
 
 var config = {
-  entry: { app: ['./src/client.js'] },
+  entry: ['babel-polyfill', './src/client.js'],
```
awong-dev: If we do this, do we need to server babel-polyfill off the produciton server? If so, where path does it need to be at? and lastly, do we need to move it from devDependencies to dependencies in package.json?

joesoftware:
LGTM
Tested on Win7-32 with IE11
`Version: 11.0.9600.17420`
`Update Versions: 11.0.14 (KB3003057)`
`Product ID: 00150-20000-00003-AA459`
hca-win7-32-ie11

robbiethegeek:
Seems like the file size of the bundle.js is growing from 6.17mb (without polyfill) to 6.9mb (with polyfill) and this is working in IE11.  I am going to suggest we don't spend more time on this unless we find this 730k is causing us any issues.  We can work on optimization post launch

awong-dev:
Make sure to measure sizes by running `npm run webpack-prod` as the default
we're using has minification off.

The size should be < 1 mb.

robbiethegeek:
with polyfill: 
bundle.js     910 kB

without polyfill: 
bundle.js     813 kB

also the npm run webpack-prod is not working as expected all it should be is NODE_ENV='"production" && webpack -p

awong-dev:
LGTM. Please file an issue for m2 or m3 for us to figure out why we need babel-polyfill.

This configuration of having both the polyfill and the transform runtime scares me from a stability perspective since the two are likely stomping on one another in unpredictable ways.

PR #95, opened 2016-05-26
Adds UI Changes from User Testing and removes redirect

crwallace:
This addresses:
- #302 Disable edit buttons for next panels on review
- #285 Review section agree to provide financial info not showing
- #303 Add green check mark for Review page
- #274 Redirecting is breaking all the things


src/client/components/HealthCareApp.jsx
```diff
@@ -32,14 +32,6 @@ class HealthCareApp extends React.Component {
     this.redirect = this.redirect.bind(this);
```
akainic: If you're removing the redirect, let's remove this too

akainic:
Can you change the name of the PR to be what you have plus "and remove redirect" so when I go back to reimplement it I can find the PR where it was removed more easily? Otherwise LGTM! When you make these couple of fixes feel free to merge.

PR #96, opened 2016-05-26
Add specs for JSON-Schema Type Definitions

markolson:
Spike on adding a way to quickly test our custom json-schema type definitions. Once the individual types are tested and stable, we can more easily move towards testing veterans.js' transformation output and the individual types, or the full schema.

Also, my first node code 🎆


test/common/schema/definitions.spec.js
```diff
@@ -0,0 +1,62 @@
+const ajv = require('ajv');
+const applicationSchema = require('../../../src/common/schema/application');
+const chai = require('chai');
+const expect = chai.expect;
+
+function definitionValidator(field) {
+  const tinySchema = {
```
awong-dev: oh interesting. clever.

test/common/schema/definitions.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    $schema: 'http://json-schema.org/draft-04/schema#',
+    type: 'object',
+    properties: {
+      field: applicationSchema.definitions[field]
+    }
+  };
+
+  return function validator(value) {
+    const fn = ajv({ allErrors: true, errorDataPath: 'property', removeAdditional: 'failing', useDefaults: true }).compile(tinySchema);
```
awong-dev: Can we yank the ajv({....}) out into a utility function in some file shared between test and prod?  I think ajv is used in 3 places now and they should all really be invoked with exactly the same parameters.

awong-dev:
LGTM with one comment.

robbiethegeek:
following up LGTM also please see if you can move the ajv invocation.

markolson:
@awong-dev @robbiethegeek 
Good call - I moved AJV into a module,

awong-dev:
LGTM!

Please resolve conflicts and then merge.

PR #98, opened 2016-05-26
Added a status API file and healthcheck endpoint

robbiethegeek:
Move the current api to a routes folder and rename application.


akainic:
LGTM, but probably want to have @alexose or @awong-dev review as well!

Now that we have this healthcheck endpoint, that's what we'll ping to make sure our app is up and running? How will that work?

robbiethegeek:
@akainic we can use this as a way for the load balancer to know that the application is up, we can also on the front end have a quick fetch of this url to validate that the node app is up and running and have that tell the end user they are one step closer to the being able to submit if the response is 200 OK.

src/server/routes/status.js
```diff
@@ -0,0 +1,12 @@
+
+function returnRouter() {
+  const router = require('express').Router(); // eslint-disable-line
+  router.get('/', (req, res) => {
+    res.status(501).end();
```
alexose: Placeholder for discoverability :)

alexose:
Seems perfectly reasonable to me.  👍

PR #99, opened 2016-05-26
Add name attributes to all inputs and build out happy path test

akainic:
Added name attributes to all inputs so that they can be targeted by nightwatch. Also built out the happy path test more. Everything is accounted for except testing for the actual submit. Submission is not currently working locally (@crwallace is working on fixing this now), so I added a TODO to complete the test once that is working.


robbiethegeek:
not reviewing but 👯

alexose:
Looks great!  I don't know if having `name` is technically best practice or not, but it seems right to me.

The test looks great and runs properly on Saucelabs:
https://travis-ci.org/department-of-veterans-affairs/healthcare-application/builds/133230018

👍

PR #105, opened 2016-05-31
Make travis build use docker for build + test.

awong-dev:
The reason for this change is llvm broke the debian apt install which
means Travis cannot easily execute npm install until someone finds
a new hosting solution for at least the llvm gpg keys.

These Dockerfiles and build scripts create an Alpine Linux based
node-4.4.4 and npm 3.8.9 container that has all the npm install
dependencies cached.

This removes most of the dependency on npm install from the Travis
build cycle in favor of one docker pull.

The idea is periodically, after any major package.json changes,
a team member would run the scripts/make-docker-build-image.sh script
to refresh the build image on the docker hub such that new travis
builds can take advantage of it.

Ideally, the build image should actually be tagged to a commit hash
that must also be placed into the travis.yml file to ensure that
it is possible to time-travel back in the build configurations, but
that is too complicated for right now and we're just trying to
make CI work again after llvm blew up npm in a fit of package
management terrorism.

Also, because the docker image now uses Alpine Linux, it does not have
a working build of phantomjs (Alpine Linux is musl based). This breaks
the default setup of Karma so as a workaround, Karma is not configured
to use saucelabs to run tests on Travis.


robbiethegeek:
Just pulled locally and verified that the local dev flow still functions and the fact that it passes on travis makes me lean towards @awong-dev thought of merge and revisit when LLVM drama is worked out.

robbiethegeek:
LGTMish :)

Dockerfile-cached-npm.build
```diff
@@ -0,0 +1,35 @@
+# This Dockerfile builds a skeletal enviornment for building and testing
```
akainic: Typo: `environment`

akainic:
LGTM (for whatever that's worth :)). Would love @alexose to test the saucelabs configuration to see what's going on there. But for now let's merge so it's not blocking the open PRs anymore.

PR #106, opened 2016-06-01
Use synchronization instead of pauses for navs.

awong-dev:
Using pause() in unittests to synchronize events (eg, waiting for a
navigation to occur after a click) is a major anti-pattern for any sort of
automated testing. Nightwatch example code in their documentation for some
reason shows this pattern but it is terrible and should not be used
because in the best case, it makes your test a lot slower than it needs
to be (most conditions seem to be meetable in tens of ms not 1000ms) and
in the worst case leads to flaky behavior because there is no stated
event to wait for.

This PR removes all calls to pause() and introduces test timeout
constants that can be adjusted globally (or maybe per environment) to
handle slower network connections in Travis, etc. Synchronization is
done by adding a little bit of instrumetnation into the top level
reaction component that allows for easy querying of the current
URL that React thinks it is rendering.

The PR also uses chai expect syntax for text expectations and only uses
nightwatch commands when directly accessing the browser/client object.
There's no great reason for this but the documentation makes it look
like the chai expect syntax is the future direction for nightwatch
so might as well use it now.

### DIFF
--- a/test/e2e/tests/00-begin.js
+++ b/test/e2e/tests/00-begin.js
@@ -6,14 +6,16 @@ const url = 'http://localhost:' + config.port; // eslint-disable-line
 
 // Test timeout constants
 // TODO(awong): Move to a common file usable for all tests.
-const defaultTimeout = 500;   // The normal timeout to use. For most opreations w/o a server roundtrip, this should be more than fast enough.
-const slowTimeout = 1000;     // A slow timeout incase the page is doing something complex.
-const molassesTimeout = 5000; // A really really slow timeout. This should rarely be used.
+const timeouts = {
+  normal: 500,     // The normal timeout to use. For most opreations w/o a server roundtrip, this should be more than fast enough.
+  slow: 1000,      // A slow timeout incase the page is doing something complex.
+  molasses: 5000,  // A really really slow timeout. This should rarely be used.
+};
 
 // TODO(awong): Move this into a custom command or assertion that can be used with client.expect.element().
-function expectNavigateAwayFrom(client, url) {
-    client.expect.element('.js-test-location').attribute('data-location')
-      .to.not.contain(url).before(defaultTimeout);
+function expectNavigateAwayFrom(client, urlSubstring) {
+  client.expect.element('.js-test-location').attribute('data-location')
+    .to.not.contain(urlSubstring).before(timeouts.normal);
 }
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


awong-dev:
This is dependent on #105 so please commit that other one first.

test/e2e/tests/00-begin.js
```diff
@@ -4,44 +4,76 @@ const config = require('../../../config');
 // FIXME: This should come in from a config variable
 const url = 'http://localhost:' + config.port; // eslint-disable-line
 
+// Test timeout constants
+// TODO(awong): Move to a common file usable for all tests.
+const defaultTimeout = 500;   // The normal timeout to use. For most opreations w/o a server roundtrip, this should be more than fast enough.
+const slowTimeout = 1000;     // A slow timeout incase the page is doing something complex.
+const molassesTimeout = 5000; // A really really slow timeout. This should rarely be used.
```
markolson: Partially because eslint complains about unused variables, would using a dict like `const timeouts = { default: 500, slow: 1000, molasses: 5000};` make sense for these?

test/e2e/tests/00-begin.js
```diff
@@ -4,44 +4,76 @@ const config = require('../../../config');
...
+    // Personal Information page.
+    client.expect.element('input[name="fname"]').to.be.visible;
+    client
       .setValue('input[name="fname"]', 'William')
       .setValue('input[name="lname"]', 'Shakespeare')
-      .click('.form-panel .usa-button-primary')
-      .pause(1000)
-      .assert.visible('select[name="veteranBirthMonth"]')
+      .click('.form-panel .usa-button-primary');
+    expectNavigateAwayFrom(client, '/veteran-information/personal-information');
```
akainic: Would it be valuable to have a similar function that expects we navigate to the next page before checking if the first input is visible?

test/e2e/tests/00-begin.js
akainic: I like that idea!

akainic:
@awong-dev Reviewed and LGTM with a couple of questions! Let me know if you want me to take on updating this branch and addressing those questions.

test/e2e/tests/00-begin.js
awong-dev: Hmm..I don't know how I feel about that.

Each stanza in my head here has a pattern of 
  (1) validate our basic environment is sane
  (2) Manipulate the page
  (3) validate the result.

Reorganizing the function to do navigate and check does reduce some code, but it also muddies up the "paragraph breaks" so-to-speak because it's combining (3) and then (1).

On the flip side, it's annoying to repeat a pattern. I could go either way...

awong-dev: done.

test/e2e/tests/00-begin.js
awong-dev: done

awong-dev:
Going to merge for now cause it'll speed up all the builds.  Let me know if you want me to change things and I'll send a follow-up PR.

test/e2e/tests/00-begin.js
akainic: Makes sense (also love the concept of stanzas). I was just thinking that validating that our basic environment is sane (#1) by checking if one particular input is visible seems incomplete. And perhaps also validating that the environment is sane by checking the url would be more complete? But I see how it's adding 2 separate checks for #1, which muddies the process.

PR #109, opened 2016-06-01
Change emailInfo type to match the VOA Spec

markolson:
Based on the `emailInfo` info in the docs, it should have a similar structure to the `phoneInfo` data


src/server/enrollment-system.js
```diff
@@ -1039,7 +1039,12 @@ function veteranToDemographicsInfo(veteran) {
           addressTypeCode: 'P',  // TODO(awong): this code is from VHA Standard Data Service (ADRDEV01) Address Type List P==Permanent. Determine if we need it.
         },
         emails: {
-          email: veteran.email,
+          email: [
+            {
+              address: veteran.email,
```
akainic: Maybe change this to `emailAddress` so it's more clear? Also this seems to be the pattern `phone` uses.

src/server/enrollment-system.js
```diff
@@ -1039,7 +1039,12 @@ function veteranToDemographicsInfo(veteran) {
...
+              type: '1'
```
akainic: Did we verify this type against the VHA standard data service?

src/server/enrollment-system.js
markolson: From https://github.com/department-of-veterans-affairs/healthcare-application/blob/master/src/server/enrollment-system.js#L991-L992 (and the doc that comes from) it looks like the key is just `address`. I may be misunderstanding how this data is supposed to be built up, though.

src/server/enrollment-system.js
markolson: I've been going off of a VOA Data Elements excel doc and the comments above this method - is there a a better place to look?

src/server/enrollment-system.js
akainic: Ah no I think you're right, ignore this!

src/server/enrollment-system.js
akainic: I think those are the right things to reference. You can also access the VHA SDS directly, but I'm not sure if you'll be able to access it outside of the VA network: http://vaausesrapp80.aac.va.gov:7404/ds/List/EmailAddressType

akainic:
LGTM! Once you rebase the branch feel free to merge.

src/server/enrollment-system.js
awong-dev: Interestingly enough, the ES system has accepted our previous submissions without this type... this is something that I might send Joshua?

markolson: Absolutely, and I'll hold off on merging until then. A concern I have is that the validator on that side is allowing extra elements through but silently dropping them, and only raising errors on malformed elements that it has a schema for.

PR #110, opened 2016-06-01
Re-vamp to match 00-begin.js

joesoftware:
Not totally happy with how I'm setting the second email address via:

`.setValue('.input-section > div:nth-child(2) input', 'bills@bard.com')`

or mobile telephone number via:

`.setValue('.input-section > div:nth-child(4) input', '3125551213')`

CSS wizards please comment.


test/e2e/tests/01-required-and-optional.js
```diff
@@ -0,0 +1,261 @@
+const report = require('../report');
+const config = require('../../../config');
+
+// FIXME: This should come in from a config variable
+const url = 'http://localhost:' + config.port; // eslint-disable-line
```
awong-dev: Pull these two into a common file?

test/e2e/tests/01-required-and-optional.js
```diff
@@ -0,0 +1,261 @@
...
+      .setValue('select[name="state"]', 'IL')
+      .setValue('input[name="zip"]', '60603')
+      .click('.form-panel .usa-button-primary');
+    expectNavigateAwayFrom(client, '/veteran-information/veteran-address');
+
+    // Contact Information Page.
+    client.expect.element('input[name="email"]').to.be.visible;
+    client
+      .setValue('input[name="email"]', 'bills@bard.com')
+      .setValue('.input-section > div:nth-child(2) input', 'bills@bard.com')
```
awong-dev: eek. counting by children under a specific page structure is indeed kinda gross.

I would instrument the code with a css class that you can select off of more easily. Either add another attribute into the Email component called `additionalInputClasses` or wrap the second sucker  a `<span>` that you can select via a class.  The `additionalInputClasses` seems the best to me though then your selector can target exactly the right thing.

test/e2e/tests/01-required-and-optional.js
```diff
@@ -0,0 +1,261 @@
...
+      .click('input[name="compensableVaServiceConnected-0"] + label')
+      .click('input[name="isVaServiceConnected-0"] + label')
+      .click('input[name="receivesVaPension-0"] + label')
+      .click('.form-panel .usa-button-primary');
+    expectNavigateAwayFrom(client, '/va-benefits/basic-information');
+
+    // Financial disclosure page.
+    client.expect.element('input[name="provideFinancialInfo-0"] + label').to.be.visible;
+    client
+      .click('input[name="provideFinancialInfo-0"] + label')
```
akainic: I wonder if for this one we should instead click `input[name="provideFinancialInfo-1"] + label` because that will expand a second question that we should also click a response for.

test/e2e/tests/01-required-and-optional.js
```diff
@@ -0,0 +1,261 @@
...
+    client
+      .setValue('input[name="insuranceName"]', 'BCBS')
+      .setValue('input[name="insurancePolicyHolderName"]', 'William Shakespeare')
+      .setValue('input[name="insurancePolicyNumber"]', '100')
+      .click('.form-panel .usa-button-primary');
+    expectNavigateAwayFrom(client, '/insurance-information/general');
+
+    // Additional VA Insurance Information Page.
+    client.expect.element('select[name="state"]').to.be.visible;
+    client
```
akainic: Let's also add the question for obtaining coverage under ACA.

akainic:
Looks good! A few fixes, and let's also update this branch with the latest from master now that we have a working submit. It's clear from the dummy data that 2 different people contributed to this :)

test/e2e/tests/01-required-and-optional.js
joesoftware: I've updated three components with an `additionalClass` (rather than `-es`) prop: ErrorableTextInput and it's children that I care about for the moment: Email and Phone.

test/e2e/tests/01-required-and-optional.js
joesoftware: Good point.

test/e2e/tests/01-required-and-optional.js
joesoftware: Missed that during the merge with Albert's changes... it's back in.

test/e2e/tests/01-required-and-optional.js
joesoftware: Not sure yet.

markolson:
`additionalClass` is a nice addition that will definitely make a lot of these UI tests more readable. 👍

akainic:
LGTM! Let's just update the branch so we can merge

PR #111, opened 2016-06-01
Update happy path test to submit to mock API endpoint

awong-dev:
Set environment variable in e2e script so that tests can post to mock API endpoint. Update existing happy path test with just required fields to submit to mock API endpoint and add a longer timeout for submission.


akainic:
@awong-dev @joesoftware I just pushed up a commit to add a longer timeout for submission only. Not that pretty, but it passes the test!

joesoftware:
LGTM

akainic:
@awong-dev Going to merge now so we can work off of this, and we can make updates later if that's okay!

PR #113, opened 2016-06-01
cleaned up package.json and recreated the npm-shrinkwrap

robbiethegeek:


awong-dev:
@robbiethegeek LGTM for b2274d3

If you're okay with 4253b7a and 4e20205 then we will have mutual LGTMs for a merge :)

jeffmaher:
Bouncing back to @robbiethegeek for review.

PR #114, opened 2016-06-02
Update the JSON-Schema spec to match the VOA docs more closely

markolson:
Quick PR to enforce a few more required fields (and field lengths) that came out of the spec docs


robbiethegeek:
LGTM

PR #115, opened 2016-06-02
Adds scrolling and validations to ReviewPanels

crwallace:


crwallace:
This closes:
316 - Scroll active review panel into view when open
265 - Add "review next" button to ReviewPanel
306 - Form Submission ID is missing on Message page

assets/css/hca.scss
```diff
@@ -236,10 +236,19 @@ button.hca-button-green {
       }
     }
     .edit-btn { 
-      padding: 1.5rem 3.5rem 1.5rem 3rem;
+      padding: 1rem;  // 1.5rem 3.5rem 1.5rem 3rem;
```
akainic: Do you want to remove the commented out styles?

src/client/components/HealthCareApp.jsx
```diff
@@ -106,10 +106,11 @@ class HealthCareApp extends React.Component {
           throw new Error(response.statusText);
         }
         response.json().then(data => {
+          console.log(data.response);
```
akainic: I think this can be removed?

src/client/components/HealthCareApp.jsx
```diff
@@ -106,10 +106,11 @@ class HealthCareApp extends React.Component {
...
           this.props.onUpdateSubmissionStatus('applicationSubmitted', data);
           this.props.onCompletedStatus(path);
-          this.props.onUpdateSubmissionId(data.formSubmissionId);
-          this.props.onUpdateSubmissionTimestamp(data.timeStamp);
+          this.props.onUpdateSubmissionId(data.response.formSubmissionId);
```
akainic: Was this to fix the id not showing up on the confirmation page? Strange that timeStamp was still being displayed

crwallace: Yes it was, and yes it was weird because it wasn't visible in the store so I am not quite sure what time it was pulling into the view.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
   handleSave() {
     const currentPath = this.props.updatePath;
-    this.props.onUpdateSaveStatus(currentPath);
+    const formData = this.props.data;
+    const sectionFields = this.props.uiData.sections[currentPath].fields;
+
+    this.props.onFieldsInitialized(sectionFields);
+    if (validations.isValidSection(currentPath, formData)) {
+      this.props.onUpdateSaveStatus(currentPath);
+      this.scrollToTop();
```
akainic: Do you need this `this.scrollToTop()` here? It should hit the one below on line 49 no matter what.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
+    }
+    this.scrollToTop();
+  }
+
+  handleNext() {
+    const currentPath = this.props.updatePath;
+    this.props.onUpdateVerifiedStatus(currentPath, true);
```
akainic: There's no "Next" button if they're in "Edit" mode, right?

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
crwallace: Nope sure don't. I think I moved it outside of the if statement and didn't delete it.

assets/css/hca.scss
```diff
@@ -236,10 +236,19 @@ button.hca-button-green {
...
+      border: 2px solid $color-primary;
+      &.primary-outline {
+        background-color: transparent;
+        color: $color-primary;
+        &:hover {
+          background-color: $color-primary;
```
akainic: I feel like this hover styling is too harsh (turning the background from transparent to that bright blue). Most of our button hover styling is to make the colors slightly darker. It's a bit harder to see for the outline buttons, but it's what we're doing elsewhere. Unless we can come up with a better solution than that, let's stick with that!

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
+    // TODO: find a better solution for this or a different implementation.
+    const delay = 100; // 0.1 second
```
akainic: So my idea of using callbacks didn't work because the `onUpdateVerifiedStatus` completes before the component renders, which is why we need the delay (one step behind you). So I think what you have is the best option for now. There's not really a need to set a variable for the delay since we're only using it one time; I think it's fine to just pass the value in directly.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
     let hiddenSection;
+    let scrollHelper;
     const currentPath = this.props.updatePath;
     const sectionsComplete = this.props.uiData.sections[currentPath].complete;
     const sectionsVerified = this.props.uiData.sections[currentPath].verified;
     const allSections = Object.keys(this.props.uiData.sections);
     const sectionIndexes = allSections.indexOf(currentPath);
     const prevPath = allSections[sectionIndexes - 1];
 
+    const ButtonEdit = (
```
akainic: Really nit picky, but this doesn't follow our casing conventions.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
+    );
 
-    if (sectionsComplete) {
-      panelAction = (<ErrorableCheckbox
-          label="I certify that all information above is correct to the best of my knowledge."
-          checked={this.props.value}
-          onValueChange={(update) => {this.props.onUpdateVerifiedStatus(currentPath, update);}}/>);
-
-      editButton = (<button
+    const ButtonNext = (
```
akainic: Same here

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
     const sectionsComplete = this.props.uiData.sections[currentPath].complete;
     const sectionsVerified = this.props.uiData.sections[currentPath].verified;
     const allSections = Object.keys(this.props.uiData.sections);
     const sectionIndexes = allSections.indexOf(currentPath);
     const prevPath = allSections[sectionIndexes - 1];
 
+    const ButtonEdit = (
+      <button
+          className="edit-btn primary-outline"
+          onClick={this.handleEdit}><i className="fa fa-pencil"></i> Edit</button>
```
akainic: Let's remove the space before `Edit`, as I think the other Edit buttons do not have that.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -1,62 +1,102 @@
...
-          label="I certify that all information above is correct to the best of my knowledge."
-          checked={this.props.value}
-          onValueChange={(update) => {this.props.onUpdateVerifiedStatus(currentPath, update);}}/>);
-
-      editButton = (<button
+    const ButtonNext = (
+      <button
           className="edit-btn"
-          onClick={this.handleEdit}>Edit Section</button>
+          onClick={this.handleNext}>Next <i className="fa fa-angle-double-right"></i></button>
```
akainic: Let's remove the space after `Next`

assets/css/hca.scss
```diff
@@ -236,10 +236,19 @@ button.hca-button-green {
       }
     }
     .edit-btn { 
-      padding: 1.5rem 3.5rem 1.5rem 3rem;
+      padding: 1rem;  // 1.5rem 3.5rem 1.5rem 3rem;
+      border: 2px solid $color-primary;
```
akainic: We're doing just a `1px` border now for the outline buttons

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -66,16 +106,18 @@ class ReviewCollapsiblePanel extends React.Component {
...
-      verifiedLabel = (
-        <div className="verify-label">
-          <span className="usa-label">&#10004; I have reviewed</span>
-        </div>);
-    }
-
-    if (sectionsVerified) {
       hiddenSection = (<div></div>);
+      buttonGroup = (<div>
+        <div className="medium-6 columns">
```
akainic: Are you using this `div` to offset the edit button so it's on the far right? You can use Foundation's offset class for that: `medium-offset-6`. Apply that to the edit button's div instead and then you can remove this one.

akainic:
Done reviewing, back to you @crwallace!

akainic:
Looks great! The only minor thing is that once we removed the space in the "Next" button on the review page there is no margin between the word "Next" and the arrow. That's because the margin applied to icons is only applied to the right. Perhaps we should change the styles on `button i.fa` to be `margin: 0 0.5em;` instead of `margin-right: 0.5em`?

crwallace:
I added classes `before-text` and `after-text` so that we only add margin on the side needed for that icon. Otherwise we get some weirdness with the button text appearing to be off center.

akainic:
@crwallace That's great! We'll also need to add those classes in `GrowableTable.jsx`.

PR #116, opened 2016-06-02
Remove npm restart and add webpack to watch.

robbiethegeek:


akainic:
LGTM except for the build fail, once that is fixed feel free to merge!

Dockerfile.travis
```diff
@@ -28,9 +28,10 @@ WORKDIR /src/healthcare-application
 # Ensure any deltas in package.json are reflected.
 COPY package.json package.json
 COPY npm-shrinkwrap.json npm-shrinkwrap.json
+COPY webpack.config.js webpack.config.js
```
awong-dev: This isn't quite right.

Node creates a layered filesystem for each command run.  You want to have the lowest amount of variance in the target filesystem so that it can get cached.

Putting this line above the `RUN npm install` line means that Docker build will not be able to reuse the node_modules of a previous build if webpack.config.js changes.

I recommend removing this line and putting the run after the `ADD . .`

That's probably the "right" way to do things.

package.json
```diff
@@ -7,10 +7,9 @@
     "karma": "karma start",
     "test:server": "mocha --recursive ./test/server ./test/common",
     "lint": "eslint --ext .js --ext .jsx lib src test",
-    "prestart": "npm run webpack-prod",
     "start": "node src/server.js",
     "test": "npm run lint && npm run test:server && npm run karma -- --single-run",
-    "watch": "npm run karma & BABEL_ENV=hot npm start & npm run watch-lint & npm run test:server",
+    "watch": "npm run karma & npm run webpack && BABEL_ENV=hot npm start & npm run watch-lint & npm run test:server",
```
awong-dev: You shouldn't need to change this line.

BABEL_ENV=hot turns on the webpack-dev-server path in server.js which runs webpack using the dev server writing to an in-memory filesystem.

What the above will do is it will run webpack once to regenerate the files in /generated on the invocation of `npm run watch` but the server will run webpack a second time for the in memory filesystem.  Then as we change source code, the in-memory filesystem will update inside the webpack-dev-server, but the on-disk version will not.

Dockerfile.travis
robbiethegeek: Moving

package.json
robbiethegeek: removed.

Dockerfile.travis
```diff
@@ -28,9 +28,10 @@ WORKDIR /src/healthcare-application
...
 
 ENV NODE_ENV development
 RUN npm install
-
+RUN npm run webpack-prod
```
awong-dev: Move this to after line 37 `ADD . .`

Then delete line 32 `COPY webpack.config.js webpack.config.js`

And it should all still work.

robbiethegeek:
Discussed with @awong-dev and merging with a minor tweak on the base docker image coming behind this.

PR #117, opened 2016-06-02
Validate enrollment-system SOAP request with the XSD

markolson:
This new test will grab the request that the soap service sends and validates it against the XSD to verify that we're sending a properly structured document over.


awong-dev:
@markolson I took the liberty of working off your branch and fixed the schema validation errors.

Also, I was wrong about the shrinkwrap problem. The libxml-xsd should have actually been added to the devDependencies (instead dependencies) which meant that shrinkwrap (as robbie most recently used it) ignores it.  Can you revert b948184?

Then feel free to squash + merge + do whatever.  It should all be green now.

awong-dev:
except...there's apparently some sort of nightwatch failure now.  Ugh.

PR #121, opened 2016-06-03
E2E test: submitting blank form fails

joesoftware:


joesoftware:
(BTW, I added classes for the submit button but didn't actually use them in this test with the hope that they'll be useful for later tests)

test/e2e/tests/02-blank-form-fails.js
```diff
@@ -0,0 +1,88 @@
...
+  'confirm-income-information',
+  'confirm-deductible-expenses',
+  'confirm-medicare-medicaid-information',
+  'confirm-insurance-information',
+  'confirm-additional-information'
+];
+
+module.exports = {
+  'Begin application': (client) => {
+    console.log(url);
```
akainic: Can we remove this `console.log`?

test/e2e/tests/02-blank-form-fails.js
```diff
@@ -0,0 +1,88 @@
...
+        value: {
+          formSubmissionId: '123fake-submission-id-567',
+          timeStamp: '2016-05-16'
+        }
+      }
+    });
+
+    // Review and Submit Page.
+    client
+      .url(url)
```
akainic: My only concern with idea of jumping directly to the review page is that we were thinking of preventing a user from being able to do that, with the idea that we want to force them to go through the form linearly because otherwise our form logic is borked. I suppose we could leave the test like this for now and then refactor if we eventually implement the redirect (I had already implemented it, but it was making it annoying to develop, so we removed it for now).

test/e2e/tests/02-blank-form-fails.js
joesoftware: Of course (and UGH I LEFT console.log in!)

test/e2e/tests/02-blank-form-fails.js
```diff
@@ -0,0 +1,88 @@
...
+  submission: 10000 // Only to be used for submission.
+};
+
+// TODO(awong): Move this into a custom command or assertion that can be used with client.expect.element().
+function expectNavigateAwayFrom(client, urlSubstring) {
+  client.expect.element('.js-test-location').attribute('data-location')
+    .to.not.contain(urlSubstring).before(timeouts.normal);
+}
+
+const section = [
```
akainic: Should the name for this be plural?

test/e2e/tests/02-blank-form-fails.js
joesoftware: Understood. Except, unless I'm not seeing things correctly, there's no way to submit a blank page via our form with page validation check preventing progress. Am I wrong?

test/e2e/tests/02-blank-form-fails.js
```diff
@@ -0,0 +1,88 @@
...
+    client
+      .url(url)
+      .waitForElementVisible('body', timeouts.normal)
+      .assert.title('Apply for Health Care: Vets.gov')
+      .waitForElementVisible('.form-panel', timeouts.slow);  // First render of React may be slow.
+
+    // confirm all sections (even though they're blank)
+    for (var i = 0; i < section.length; i++) { // eslint-disable-line
+      client
+        .waitForElementPresent(`span.${section[i]} button`, timeouts.molasses)
```
akainic: Is the `timeout.molasses` necessary for the sections to appear? I'm surprised they take that long to render :(

src/client/components/HealthCareApp.jsx
```diff
@@ -158,20 +158,23 @@ class HealthCareApp extends React.Component {
     if (submissionStatus === false) {
       submitButton = (
         <ProgressButton
+            additionalClass="submit-status-not-submitted"
```
akainic: Did you add this in order to figure out what "state" the button is in for testing? You should be able to do that with Nightwatch's `text` expectation and use that to see what the text of the button is, since that will indicate the state.

test/e2e/tests/02-blank-form-fails.js
akainic: You're totally right, I was thinking about that earlier. I guess the only other way we could do it is to try to submit each section blank, expect not to progress, then complete fields to get to the next section and try it again.

src/client/components/HealthCareApp.jsx
joesoftware: That's exactly the case. If this has no value, we can remove in lieu of `text` checking. I got class crazy last night.

test/e2e/tests/02-blank-form-fails.js
joesoftware: Nope, can change to different duration.

test/e2e/tests/02-blank-form-fails.js
joesoftware: Probably... I copied from elsewhere and left the plurality as original. I never read code as english anymore b/c I'm old and seen too much code where the names are short or nonsensical.

src/client/components/HealthCareApp.jsx
akainic: Gotcha. I think adding additional classes should be a last resort, so let's see if we can target these instead by checking the text of the button.

src/client/components/ProgressButton.jsx
```diff
@@ -22,10 +24,12 @@ class ProgressButton extends React.Component {
     const afterText = (this.props.afterText) ? (<span className="button-icon"> {this.props.afterText}</span>) : '';
 
     return (
-      <button
-          className={`text-capitalize ${this.props.buttonClass}`}
-          id={`${this.id}-continueButton`}
-          onClick={this.props.onButtonClick}>{beforeText}{this.props.buttonText}{afterText}</button>
+      <span className={this.props.additionalClass}>
```
akainic: I think these changes can be reverted too if we're removing the `additionalClass` prop on the buttons.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
```diff
@@ -52,16 +52,21 @@ class ReviewCollapsiblePanel extends React.Component {
       panelAction = (<ErrorableCheckbox
           label="I certify that all information above is correct to the best of my knowledge."
           checked={this.props.value}
+          additionalClass={this.props.additionalClass}
```
akainic: FYI this is going to change because we're removing the checkboxes and replacing them with "Next" buttons.

test/e2e/tests/02-blank-form-fails.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // Review and Submit Page.
+    client
+      .url(url)
+      .waitForElementVisible('body', timeouts.normal)
+      .assert.title('Apply for Health Care: Vets.gov')
+      .waitForElementVisible('.form-panel', timeouts.slow);  // First render of React may be slow.
+
+    // confirm all sections (even though they're blank)
+    for (var i = 0; i < section.length; i++) { // eslint-disable-line
```
akainic: I'm a little confused by what this is doing. I think you can submit the form without updating each of the sections or clicking that you verify anything. In fact, if you click update for a blank section, it shouldn't allow you to proceed to the next one because it is invalid. Is this necessary to do for the purpose of the test?

joesoftware: It felt more complete to "confirm" each section. Ultimately this is a balance between knowing how things work internally and externally testing things. The more I think about this test case the more meaningless it seems...

src/client/components/ProgressButton.jsx
joesoftware: Agreed.

src/client/components/form-elements/ReviewCollapsiblePanel.jsx
joesoftware: Let's leave this until that's in place.

src/client/components/ReviewAndSubmitSection.jsx
```diff
@@ -47,81 +47,97 @@ class ReviewAndSubmitSection extends React.Component {
 
         {/* TODO(crew): Change names of sections to real names. */}
         <ReviewCollapsiblePanel
+            additionalClass="confirm-personal-information"
```
awong-dev: nit: please prefix css classes that are used for non-style reasons by JS with `js-`

test/e2e/tests/02-blank-form-fails.js
awong-dev: This might be an indication that you're attempting to use an e2e test to do something a unittest is actually meant for.

What you seem to be trying to assert is "if redux store is blank and someone somehow manages to render the submit button and cilck it, we don't let it happen".  In that world, I'd want to just render the submit component using the Karma tests, manually set the store to something useless, and then test all those reactions.

Using an e2e test means your test "setup" becomes heavy weight for the assertion you are trying to make...

joesoftware: Perhaps we should re-classify our tests with this in mind?

akainic: I think we should still have an e2e test that tries to submit each section without all of the required fields (or none) and makes sure that it cannot be submitted.

joesoftware:
Changes to master make most of this PR irrelevant. Discussing next steps.

akainic:
@joesoftware I'm going to close this PR since it seems like it is now obsolete.

PR #123, opened 2016-06-03
WIP: Fields initialized test

akainic:


markolson:
I like it! The main thing I'd think about adding would be adding a second parameter of values to set the fields as (that could be merged in with some defaults), so that the same functions could be used to test the error handling when we get off the happy path.

joesoftware:
I was thinking the same thing that Mark said... maybe an optional object in field/value form for overrides or additions to fields?

akainic:
Great ideas @markolson and @joesoftware! I will incorporate that.

akainic:
@markolson If you wouldn't mind taking a second look at this, I added some test data that we pass to the completion functions (right now only the first one, `completePersonalInformation` is refactor to accept it). Curious to get your thoughts if this is along the lines of what you were thinking!

test/e2e/utils/common.js
```diff
@@ -6,10 +6,485 @@ const timeouts = {
...
+    field: 'input[name="campLejeune"] + label'
+  }
+};
+
+function completePersonalInformation(client, data, onlyRequiredFields) {
   client
-    .setValue('input[name="fname"]', 'William')
-    .setValue('input[name="lname"]', 'Shakespeare');
+    .setValue(data.veteranFullName.first.field, data.veteranFullName.first.value)
+    .setValue(data.veteranFullName.last.field, data.veteranFullName.last.value);
```
markolson: I think that defining the `field` attribute may be half a step too far - I'd expect those to be consistent between the tests, no matter the `value`.

markolson:
Besides maybe simplifying the input data to just be `fieldName: "data"` rather than `fieldName: {value: "data"}`, I like the latest change, especially the requiredOnly flag

joesoftware:
LGTM with a comment. Due to no fault of this PR, the test `01-required-and-optional.js` does not check the form success page.

PR #124, opened 2016-06-06
Removes children/providers data if unchecked

crwallace:
Closes: 
298 - Delete children/providers if unchecked


akainic:
LGTM! Let's just add a TODO to look into why children/providers data isn't getting updated in the store as it's being entered.

PR #127, opened 2016-06-08
Make a place for questions and answers.

joesoftware:


markolson:
My initial thought is that this may be over-specific, but having the baseline may be handy if anyone else hits this problem.. I think the more important thing to include is how to _get_ to that baseline, especially if something is wrong.

akainic:
I agree with @markolson that this seems very specific. Perhaps instead something simply saying, "If you encounter this problem, it may be because your development environment needs to have the following version of xcode", with "this problem" being the problem you encountered.

joesoftware:
Reframed this as suggested.

markolson:
I'd maybe move it to the docs/ folder, then 👍 from me

joesoftware:
Sounds good Mark. I'll move the file, update the branch then merge with master.

PR #128, opened 2016-06-08
Update waffle 309 (E2E for 01-required-and-optional.js)

joesoftware:


akainic:
This looks great! I think the other thing we need to do for this test is refactor it to use the completion functions. I think we are using them for the first section, but none of the others. If you want to take care of that in a separate PR that's fine too! The work that's here is ready to merge.

PR #129, opened 2016-06-08
Convert email addresses to lower case before comparing

akainic:
Currently the comparison between `email` and `emailConfirmation` is case sensitive, which it should not be. Convert both email addresses to lower case before comparing so they will not be case sensitive.


crwallace:
LGTM!

PR #130, opened 2016-06-09
Remove requirement for phone

akainic:
Phone number fields were being indicated as required, though they are not. Remove required prop.


markolson:
LGTM after an update/rebase

PR #131, opened 2016-06-09
Adds tooltip functionality

crwallace:
closes: 
- 245: Add Tooltip component

screen shot 2016-06-09 at 1 47 28 pm

**NOTE: This copy is dummy copy.**

### DIFF
--- a/src/client/components/form-elements/DateInput.jsx
+++ b/src/client/components/form-elements/DateInput.jsx
@@ -4,6 +4,8 @@ import _ from 'lodash';
 import ErrorableSelect from '../form-elements/ErrorableSelect';
 import ErrorableNumberInput from '../form-elements/ErrorableNumberInput';
 
+import ToolTip from '../form-elements/ToolTip';
+
 import { validateIfDirtyDate, isBlank, isValidDate } from '../../utils/validations';
 import { months, days } from '../../utils/options-for-select.js';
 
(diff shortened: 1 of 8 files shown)


src/client/components/form-elements/ToolTip.jsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  render() {
+    return (
+      <div className="hca-tooltip">
+        <button className="help-link" tabIndex={this.props.tabIndex} onClick={this.handleOpen}>More Info</button>
+        <div className="help-content call-out" aria-hidden={this.state.closed}>
+          {this.props.toolTipText}
+          <br/>
+          <button className="close-link" tabIndex={this.props.tabIndex} onClick={this.handleClose}>Close</button>
```
akainic: Doesn't look this `classname="close-link"` is being used for anything, can we remove it?

src/client/components/form-elements/DateInput.jsx
```diff
@@ -68,6 +72,16 @@ class DateInput extends React.Component {
       errorSpan = <span className="usa-input-error-message" id={`${errorSpanId}`}>{this.props.errorMessage}</span>;
     }
 
+    // Addes ToolTip if text is provided.
```
akainic: Typo: `Adds`

akainic:
Couple of thoughts:
1.) Can you add a comment in `ToolTip` that explains why taborder is needed and how to apply it to the tooltip? I imagine this could get tricky once we need to specify taborder for other components to fix issue #296 on tabbing, so it could be helpful to have a little description of how to use it properly for the tooltip.
2.) I'm wondering if it would be better functionality if the `open` and `close` link were the same, and the text itself changed based on the state of the tooltip. And the functionality just toggled the state. So if you click the link, it checks what the current state is and it sets it to the opposite? Wondering if that's better than having 2 separate links.

crwallace:
1. Sounds good! I can take care of this. 
2. I only created two links because that was the suggestion for making a properly accessible tooltip. Let me take a deeper look and see why it recommends two links.

akainic:
LGTM!

PR #133, opened 2016-06-09
354 validations breaking submissions

robbiethegeek:
- Minor tweaks to our processing of home phone and mobile phone for veteran’s 
  -TODO to revisit maiden name

### DIFF
--- a/src/server/enrollment-system.js
+++ b/src/server/enrollment-system.js
@@ -88,14 +88,12 @@ function spanishHispanicToSDSCode(isSpanishHispanicLatino) {
 }
 
 /**
- * Returns SDS code representing the indication of Spanish/Hispanic/Latino ethnicity.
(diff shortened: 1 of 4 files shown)


src/server/enrollment-system.js
```diff
@@ -88,6 +88,37 @@ function spanishHispanicToSDSCode(isSpanishHispanicLatino) {
 }
 
 /**
+ * Returns SDS code representing the indication of Spanish/Hispanic/Latino ethnicity.
```
awong-dev: This comment looks wrong.  We're returning phone numbers?

awong-dev:
LGTM after fixing comment to match.

src/server/enrollment-system.js
robbiethegeek: Haha copy paste of a method above to save writing so little. Will update then merge

PR #134, opened 2016-06-10
Add a Version to the ES SOAP request

markolson:
This field is unused in ES and can, for a time, be used as a way to track which submissions come from this application.

Closes Waffle #358

### DIFF
--- a/src/server/enrollment-system.js
+++ b/src/server/enrollment-system.js
@@ -24,7 +24,10 @@ const formTemplate = {
     //  * form / formIdentifier / type, Not Applicable, Yes, "Data element is not a form captured element but provides the type of form submitted  (e.g., . 1010EZ)"
     formIdentifier: {
       type: '100',
-      value: '1010EZ'
+      value: '1010EZ',
+      // Version can act as a marker in the XML stored in ES that this request came from vets.gov
+      // "vets".each_byte.map{|b| b.to_s(16) }.join.to_i(16)
+      version: 1986360435
     },
     //  * identity / authenticationLevel / type, Not Applicable, No, Data element is not a form captured element but provides the  level of the Veterans' authentication.
     //  * identity / veteranIdentifier / type, Not Applicable, Yes, "Data element is not a form captured element but provides the Veteran Identifer type (e.g,  EDIPI)"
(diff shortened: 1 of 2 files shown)


markolson:
For posterity, if we need to add a comment to the generated XML, it looks something like..

```
/* If we pass a string into the saveSubmitForm client method, it will be passed through.
   To do that, we duplicate the logic used to create it, and then quickly splice in our
   comment node that will be logged on the ES side. This is.. not ideal, but it's the
   best I've come up with while still using this library. */
// get the wsdl definition object from the service
const definition = voaService.wsdl.definitions.services.voaService.ports.VoaFormServiceEndpointPort.binding.methods.saveSubmitForm;
// dig into the input type 
const inputType = definition.input;
// with the input type, construct the XML document that will be the SOAP body. This is the same call done internally if a JSON object is provided to the soap service method.
const xml = voaService.wsdl.objectToDocumentXML(inputType.$name, saveSubmitFormMsg, inputType.targetNSAlias, inputType.targetNamespace, (inputType.$type || inputType.$lookupType));
// inject a tracer comment node so that ES can see which enrollments come from this system
const outputXML = xml.replace('<va:formIdentifier>', '<va:formIdentifier><!-- VETS.GOV -->');
voaService.saveSubmitForm(outputXML, (err, response) => { 
....
```

akainic:
Looks good to me. Would love @robbiethegeek's eyes on this as well.

src/server/enrollment-system.js
```diff
@@ -24,7 +24,10 @@ const formTemplate = {
     //  * form / formIdentifier / type, Not Applicable, Yes, "Data element is not a form captured element but provides the type of form submitted  (e.g., . 1010EZ)"
     formIdentifier: {
       type: '100',
-      value: '1010EZ'
+      value: '1010EZ',
+      // Version can act as a marker in the XML stored in ES that this request came from vets.gov
+      // "vets".each_byte.map{|b| b.to_s(16) }.join.to_i(16)
+      version: 1986360435
```
robbiethegeek: clever 👍 thanks for documenting how you got to this.

robbiethegeek:
When the tests pass feel free to merge @markolson LGTM!

PR #135, opened 2016-06-10
Add e2e test for final submit button text

austin-martinez:


markolson:
Is this mostly to test the `✓ Submitted` value? If so, we may want to take a step back and see if we can push even more of the form-filling logic into some module or class sooner rather than later.

akainic:
LGTM! I agree with @markolson on that. Let's create a separate ticket for that and prioritize it over the next e2e tests? I think we may still want to merge this for now as is and refactor as part of extracting out the shared logic into a separate module or class. Just so we're not holding up this PR.

akainic:
Created a ticket for that: https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/363.

robbiethegeek:
LGTM also restarted the tests will merge if all passes :)

PR #136, opened 2016-06-13
Add additional options

akainic:
Received a response back from Mehedi confirming that we can use additional options. Closes this ticket: https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/291.


crwallace:
LGTM!

PR #142, opened 2016-06-14
E2E testing that the form can't advance without required fields

markolson:
This is basically 00-required-fields, just with invalid submissions at each step

### DIFF
--- a/test/e2e/tests/04-cannot-progress-without-required-fields.js
+++ b/test/e2e/tests/04-cannot-progress-without-required-fields.js
@@ -1,0 +1,188 @@
+const request = require('request');
+
+const report = require('../report');
+const config = require('../../../config');
+
+// FIXME: This should come in from a config variable
+const url = 'http://localhost:' + config.port; // eslint-disable-line
+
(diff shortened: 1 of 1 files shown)


akainic:
Looks good! Not sure why that one test is failing, it seems correct. Maybe try restarting the build and see what happens?

Also, I see for the first test you fill out 1 field after filling out none and verify that that also does not allow the form to be submitted, but we don't follow this pattern for the other tests. I'm torn on this. It seems awfully tedious to do this for all sections, so we might as well be consistent with the pattern and remove this in the first section?

markolson:
For the failures, it's the same thing that's happening with me locally, which.. is not great. It looks like the submit button isn't being hit (or triggering it's action), and not advancing the page. So, the failure is legit, if intermittent. I'm still investigating this to see _where_ the problem ultimately lies, but I don't want to merge until it's stable.

For the more thorough testing, I was also torn on going through and testing a field or two at a time on more than just the first section, but thought that having a small smoke test in one of them could be worthwhile.

PR #143, opened 2016-06-14
Remove spouseData from submission if there is no spouse

markolson:
I'm sure there's a cleaner way to do this in the React component itself, but putting the logic in with the rest of the scrubbing in `veteranToApplication` seemed like a safe way to start.


robbiethegeek:
can you move this to using lodash : https://lodash.com/docs#includes
if (_.includes(['Never Married', 'Widowed', 'Divorced'], veteran.maritalStatus.value))

markolson:
TIL. Updating.

robbiethegeek:
LGTM

PR #145, opened 2016-06-15
Remove provideFinancialInfo question and fix spouse SSN validation

akainic:
Remove `provideFinancialInfo` question because it does not map to anything in ES or in the paper application. Also fix spouse SSN validation: section validation function was allowing blank SSN values to pass validations, but this is incorrect.

### DIFF
--- a/src/client/components/household-information/FinancialDisclosureSection.jsx
+++ b/src/client/components/household-information/FinancialDisclosureSection.jsx
@@ -14,38 +14,15 @@ import { veteranUpdateField } from '../../actions';
 class FinancialDisclosureSection extends React.Component {
   render() {
     let content;
-    let understandsFinancialDisclosure;
-    let understandsFinancialDisclosureReview;
-
-    if (this.props.data.provideFinancialInfo.value === 'N') {
-      understandsFinancialDisclosure = (<div>
-        <ErrorableRadioButtons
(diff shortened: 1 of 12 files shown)


src/client/components/household-information/FinancialDisclosureSection.jsx
```diff
@@ -14,38 +14,15 @@ import { veteranUpdateField } from '../../actions';
...
 
     if (this.props.isSectionComplete && this.props.reviewSection) {
       content = (<table className="review usa-table-borderless">
         <tbody>
           <tr>
-            <td>I agree to provide my financial information so the VA can determine my eligibility
-             for VA healthcare and if I should be charged for copays and medication.:
-            </td>
-            <td>{`${this.props.data.provideFinancialInfo.value === 'Y' ? 'Yes' : 'No'}`}</td>
+            <td>I understand VA is not currently enrolling new applicants who decline to
```
robbiethegeek: is this the new verbage from the lawyers / stacey?

akainic: No, we will wait to update that language until we get the full list of changes from her, which I have not received yet.

robbiethegeek:
LGTM one question about the verbage if the answer is yes and the tests pass we can / should merge.

PR #149, opened 2016-06-16
Section validation for birth info was not catching invalid SSN entries

akainic:
Logic is hard


robbiethegeek:
LGTM!   Once tests confirm let's merge.

PR #151, opened 2016-06-16
Change email serialization in SOAP

markolson:
- Leave the emails/email nodes out entirely if no emails are provided
- Move the `email` into an array of `emails`, though we only have one for now

Closes #382


robbiethegeek:
LGTM once security is resolved or exempted we can retest / merge

PR #152, opened 2016-06-16
Update expresss and fix broken e2e tests

robbiethegeek:


robbiethegeek:
@plusjeff this is the static analysis PR that was trigger by the output of npm run security

akainic:
So interesting that removing the label now makes the `click` assertion work. I remember that distinctly not working when I was first working on these tests, not sure if the express upgrade changed that, but good find!! LGTM!

PR #154, opened 2016-06-17
Legal changes

crwallace:
This includes all feedback from the legal team.


src/client/components/household-information/AnnualIncomeSection.jsx
```diff
@@ -43,15 +43,15 @@ class AnnualIncomeSection extends React.Component {
...
                   <td>{child.grossIncome.value}</td>
                 </tr>
                 <tr>
-                  <td>Children Net Income:</td>
+                  <td>Net Income from your Farm, Ranch, Property or Business:</td>
                   <td>{child.netIncome.value}</td>
                 </tr>
                 <tr>
-                  <td>Children Other Income:</td>
+                  <td>Other Income AmountS:</td>
```
akainic: Typo I think: capital `S` at the end

src/client/components/household-information/AnnualIncomeSection.jsx
```diff
@@ -43,15 +43,15 @@ class AnnualIncomeSection extends React.Component {
             <table className="review usa-table-borderless">
               <tbody>
                 <tr>
-                  <td>Children Gross Income:</td>
+                  <td>Gross Annual Income from Employment:</td>
```
akainic: Do we want to remove `Child` from the beginning? Seems like that could be useful in being super clear about what each field maps to

src/client/components/ReviewAndSubmitSection.jsx
```diff
@@ -131,6 +131,8 @@ class ReviewAndSubmitSection extends React.Component {
...
...is responsible for payment of the charges for my medical care, including benefits otherwise payable to me or my spouse. Furthermore, I hereby assign to the VA any claim I may have against any person or entity who is or may be legally responsible for the payment of the cost of medical services provided to me by the VA. I understand that this assignment shall not limit or prejudice my right to recover for my own benefit any amount in excess of the cost of medical services provided to me by the VA or any other amount to which I may be entitled. I hereby appoint the Attorney General of the United States and the Secretary of Veterans' Affairs and their designees as my Attorneys-in-fact to take all necessary and appropriate actions in order to recover and receive all or part of the amount herein assigned. I hereby authorize the VA to disclose, to my attorney and to any third party or administrative agency who may be responsible for payment of the cost of medical services provided to me, information from my medical records as necessary to verify my claim. Further, I hereby authorize any such third party or administrative agency to disclose to the VA any information regarding my claim.</p>
```
akainic: This makes me sad :( But I understand the need for it. Can we at least make this font a few px smaller?

src/client/components/household-information/AnnualIncomeSection.jsx
```diff
@@ -67,21 +67,21 @@ class AnnualIncomeSection extends React.Component {
           <h6>Spouse</h6>
           <ErrorableTextInput
               errorMessage={getErrorMessage(this.props.data.spouseGrossIncome, message)}
-              label="Spouse Gross Income"
+              label="Gross Annual Income from Employment"
```
akainic: I feel like it's still helpful to preface each of the labels with `Veteran`, `Spouse`, or `Child`. So it would read: `Spouse Gross Annual Income from Employment`. Just because otherwise the labels are all the same, and, especially from an accessibility perspective, that could be confusing.

src/client/components/household-information/AnnualIncomeSection.jsx
crwallace: Yesterday we said that we were going to leave the section titles "Veteran", "Spouse", "Child: Name" and all of the input labels would be the same.

akainic: Oh oops, must have missed/forgotten that decision. I'm just worried about the case of a screenreader if it will be clear that a field is associated with `Spouse`, for example, if it's not in the label text.

akainic:
LGTM!

PR #157, opened 2016-06-17
Don't include SpouseFinancials if there is no financial data

markolson:
Applies similar logic to the dependents' financial info

Closes #387


robbiethegeek:
LGTM

PR #158, opened 2016-06-17
Update production 061716 2

akainic:


robbiethegeek:
I still think this code LGTM :) 👍  once tests pass let's trigger a prod deploy

PR #162, opened 2016-06-20
adds character limits to inputs to pass VOA validation

crwallace:
closes:
- 392: Add Max 30 characters
- 393: Birth City max 20 characters
- 394: Email max 50 characters


akainic:
@crwallace Can you include a screenshot of what the new max character message looks like?

src/client/components/form-elements/ErrorableTextInput.jsx
```diff
@@ -66,6 +67,10 @@ class ErrorableTextInput extends React.Component {
       requiredSpan = <span className="hca-required-span">*</span>;
     }
 
+    if (this.props.charMax) {
+      maxCharacters = (<small>(Max. {this.props.charMax} characters)</small>);
```
akainic: I would just spell out `Maximum` to be more clear

akainic: I think it might be better to only have this message appear if they reach 30 characters. And also perhaps have it be part of the error message.

src/client/components/veteran-information/BirthInformationSection.jsx
```diff
@@ -55,6 +55,7 @@ class BirthInformationSection extends React.Component {
               name="cityOfBirth"
               field={this.props.data.cityOfBirth}
               autocomplete="off"
+              charMax={20}
```
akainic: Just want to verify that this has to be 20 max characters, but city in `Address` can be 30?

markolson: From the spreadsheet I've been looking at, 'Place of Birth' city can be up to 20, 'Current' City can be up to 51, 'Next of Kin' city can be up to 30, 'Emergency Contact' can be 30, 'Insurance Company' city can be 25, 'Employment' city can be 30. Cities are just all over the place.

akainic: Ah gotcha. Thanks for that info!

crwallace:
This is what character max will look like. 
screen shot 2016-06-21 at 10 15 31 am

@maryannbrody and I worked on this design yesterday and we thought "character" needed to be spelled out but not "max".

src/client/components/form-elements/ErrorableTextInput.jsx
crwallace: I was trying to prevent us from doing two things: 1. giving the user an error message that contains more than one thing in it, and 2. adding in more front-end validations when there is an html attribute that does this already.

akainic: I don't think it's a problem to have an error message say: "Please enter a valid name. Maximum 30 characters," or something like that. To me, it's more confusing to have validation feedback in 2 different formats: as an error message and as help text below the field. It's also unnecessary in the vast majority of cases to show this message, since most probably won't hit this max. I think it would be better to only show this message, then, if they hit the max.

I'm totally fine with using the html attribute instead of our js validation functions for this. I think it's more of a question of how your provide feedback on the fact that there is a maximum number of characters allowed. One way you could do this is:
```
if (this.props.errorMessage) {
    errorSpanId = `${this.inputId}-error-message`;
    errorSpan = <span className="usa-input-error-message" id={`${errorSpanId}`}>{this.props.errorMessage}</span>;
    inputErrorClass = 'usa-input-error';
    labelErrorClass = 'usa-input-error-label';
} else if (this.props.field.value.length > this.props.charMax) {
    errorSpanId = `${this.inputId}-error-message`;
    errorSpan = <span className="usa-input-error-message" id={`${errorSpanId}`}>Maximum 30 characters</span>;
    inputErrorClass = 'usa-input-error';
    labelErrorClass = 'usa-input-error-label';
}
```

src/client/components/form-elements/ErrorableTextInput.jsx
```diff
@@ -43,9 +43,9 @@ class ErrorableTextInput extends React.Component {
     let errorSpanId = undefined;
     let inputErrorClass = undefined;
     let labelErrorClass = undefined;
-    if (this.props.errorMessage) {
+    if (this.props.errorMessage || this.props.field.value.length > this.props.charMax) {
```
akainic: Hm I didn't really think this through... I liked how you used the `maxLength` attribute, but I didn't realize that the message will never show up if we use it, hence why you removed that. Sorry! I'm rethinking this now. Perhaps the small text you had below the field is the best way to go, but just show it conditionally if `this.props.field.value.length >= this.props.charMax` so it'll only show up once they hit 30 characters, for example. Thoughts on that?

crwallace: Are you still wanting to use `maxLength`?

akainic: @crwallace I totally agree. Let's use `maxLength`. Can we still show that message below the input (what you originally had) conditionally, though, on `this.props.field.value.length >= this.props.charMax`? I don't think that will cause a test failure if we first check if `this.props.field.value` is not blank?

akainic:
LGTM!

PR #166, opened 2016-06-22
Update production 062216

akainic:


robbiethegeek:
LGTM once the tests fast I think we are good

PR #167, opened 2016-06-22
Changes wantsInitialVaContact from checkbox to radio button.

crwallace:
This just changes one checkbox to a radio button as a template.


akainic:
Looks great! A couple of things to fix:
1. Looks like you have a lint error
2. I think we need to change the review section part for this question to: `this.props.data.wantsInitialVaContact.value === 'Y' ? 'Yes' : 'No'`

Also, did we try testing this to make sure the submit still works?

akainic:
LGTM!

PR #174, opened 2016-06-24
Always include provided Spouse and Dependent demographic info

markolson:
- `incomes` will still be left out of there's no income data
- demographic data will always be included if present
- `discloseFinancialInformation` will only be included if incomes, expenses, spouse income, or dependent income is set.


robbiethegeek:
This looks pretty clean and well thought out @markolson once the tests pass I will give it one more pass then LGTM likely and merge,

@joduinn looping you in for clarity

PR #177, opened 2016-06-24
Update country names

austin-martinez:
Used country names from here
http://www.worldatlas.com/aatlas/ctycodes.htm

Do we need to update the ordering? Wasn't sure if they were in alpha order by code or name?


crwallace:
I think we got these codes from the ES so we should check and make sure that we didn't add any new ones because I don't know if the ES will have a code for them and I don't want it to error out for that weird edge case.

austin-martinez:
the codes didn't change, just the names/labels

crwallace:
LGTM!

PR #179, opened 2016-06-24
Fix search button mobile

alexose:
This PR fixes https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/402 by means of updating the header and footer to what's in vets-website.

I'm not thrilled about having to pull in jQuery, but reimplementing the vets-website menu in react probably isn't a good use of time.

IMO the best course of action is to prod the vets-website folks towards a minimal, dependency-free header and footer that can be consumed by any frontend stack.  While I'm not going to light that fire just yet, it will probably happen as a result of creating a component library.


alexose:
Blah, not sure why checks are failing... looking into it

assets/css/va.scss
```diff
@@ -2189,103 +2189,358 @@ button[class*="usa-button-"] {
...
-  	}
+    .overlay form {
+      height: 70%;
+    }
+  }
+}
+
+// TODO: Remove !important when .footer a selector gets simplified
+
+.va-crisis-panel {
```
akainic: Where did these styles come from? I don't see them in _va.scss in vets-website.

public/index.html
```diff
@@ -224,25 +237,59 @@ <h2 class="sr-only">Apply for Health Care</h2>
...
+
+UserVoice=window.UserVoice||[];(function(){var uv=document.createElement('script');uv.type='text/javascript';uv.async=true;uv.src='//widget.uservoice.com/NDkEKHU7ZuU64RQKkUL9Q.js';var s=document.getElementsByTagName('script')[0];s.parentNode.insertBefore(uv,s)})();
+
+UserVoice.push(['identify', {
+}]);
+
+UserVoice.push(['autoprompt', {}]);
+</script>
+
+<!-- Google Tag Manager -->
```
akainic: We already have a Google Tag Manager tag earlier in the file, so I don't think we need this one.

akainic:
@alexose I agree about pushing vets-website towards using a dependency-free header and footer, and that our best solution is to use this for now and improve post-launch. Let's file an issue in the vets-website repo about this to keep track of for later.

akainic:
@alexose, I tested this locally and it works great! Not sure why the tests are failing, but let's address the comments above, update the branch, and restart Travis to see if that fixes things. If not, we can look more into why some tests are failing.

public/index.html
alexose: Oops, my bad.

assets/css/va.scss
alexose: This are via `_sass/components/_va-crisis-line.sc`... @webinista recommended I add them

akainic: Oh sorry, I didn't mind having it in, just wanted to know where it came from! I assume we want to include the Crisis Line code?

alexose:
Seems to be passing now. @akainic, can I get a review/merge?

akainic:
Hey @alexose! I tested this locally and the Crisis Line stuff in the footer looks a little wonky. I think we should include the new Crisis Line css and html, my comment earlier was more to understand where it was coming from, not that we shouldn't include it. Sorry for the confusion!

alexose:
Turns out I'd gotten a little overzealous with deleting CSS.  Looks better now, although it's a bit off-brand from the one in vets.gov.  This will be addressed once we unify the vets.gov assets.

image

@akainic Once this passes tests, it should be ready to go!

akainic:
LGTM!

PR #182, opened 2016-06-27
Initial json->xml conformance maps for financial details

markolson:
These static files exercise the logic in enrollment-system to make sure that Financial data is serialzied into XML correctly given different responses. I'm hoping that while large and complex, they can help us stick a pin in what the output is supposed to look like. 

`child-financial` is when the vet doesn't have info, but children do.
`spouse-financial` is when the vet doesn't have info, but spouse does.
`no-financial` is when no finanial data is provided.

Child & Spouse tests should (and do) set the discloseFinancialInformation flag, while no-financial shouldn't, though the demographic data should be provided in all cases.


markolson:
As a note, I went with full files for the JSON input because munging the Golden JSON with code ended up getting harder to follow, and was somewhat brittle in the face of that file changing.

test/server/enrollment-system.spec.js
```diff
@@ -36,6 +36,24 @@ describe('enrollment-system base tests', () => {
       result.should.deep.equal(goldenJsonSubmission);
     });
 
+
+    const checks = ['child-financial', 'no-financial', 'spouse-financial'];
+    for (const filename of checks) {
+      console.log(filename);
```
akainic: Do you want to remove the `console.log`?

markolson: Yes. Yes I do.

akainic:
This looks great and is a great idea for us to include. How did you generate the xml files?

markolson:
XML was copied from a `console.log(messageBody)` in the `client.on('message', (messageBody) => {` callback

akainic:
Great, LGTM!

PR #183, opened 2016-06-27
Update to the latest facility list.

robbiethegeek:
Closes ticket #301


crwallace:
There are still some duplicates in this list. Only looked quickly but under FL I still see two `'ORLANDO VA HEALTHCARE CENTER NHCU'`.

crwallace:
Also I just noticed this as well `{ value: '673PA', label: 'THIS ENTRY ENTERED IN ERROR' },`

robbiethegeek:
@akainic @staceylanger not sure how you want to handle this?  I think there may be dupes but they have different codes and then what about the error'd ones?

crwallace:
Also...looks like some of these might not be medical facilities/hospitals `{ value: '5099AA', label: 'AUGUSTA NURSING HOME' },`

akainic:
@robbiethegeek and @crwallace sounds like a question for Mehedi. I will follow up with him on this.

crwallace:
This looks much better!

akainic:
LGTM! Will merge once tests pass.

PR #184, opened 2016-06-27
297 checkbox to radio

austin-martinez:


crwallace:
@akainic are these the only questions we wanted to change to radio buttons? I know that there are other questions like "I am enrolling to obtain minimal essential coverage under the affordable care act" that we want to change from checkbox to radio button.

akainic:
I think we also need to change references to these fields in `test/e2e/utils/common.js`.

akainic:
@crwallace I went through the form and the child fields + the ACA question were the only remaining checkboxes that aren't lists of things (like military service history and demographic information). I thought we decided to leave the ACA question as a checkbox because it is written as an "I" statement, so it doesn't really make sense to make it a radio button without changing the way the question is worded.

crwallace:
@akainic someone brought it up on the demo call on Friday and I thought it was part of this change so I didn't make a new ticket. Sorry about that. Yeah we should probably add it to the list.

test/e2e/utils/common.js
```diff
@@ -105,11 +105,11 @@ const testValues = {
...
       },
-      childDisabledBefore18: false,
-      childAttendedSchoolLastYear: false,
+      childDisabledBefore18: '',
+      childAttendedSchoolLastYear: '',
       childEducationExpenses: '6000',
-      childCohabitedLastYear: false,
-      childReceivedSupportLastYear: false,
+      childCohabitedLastYear: '',
+      childReceivedSupportLastYear: '',
```
akainic: I think this file needs to be updated on lines 361 and onward, for the `completeChildInformation` section on what inputs to click.

akainic:
LGTM! Let's leave changing that additional question to a radio button for another PR. I will create a ticket for it.

PR #187, opened 2016-06-28
Add build version to status endpoint.

awong-dev:


robbiethegeek:
The test for the / of status is expecting a 501 return not a 200.  @awong-dev let's chat as to what is the need here.

awong-dev:
There's no need. this was just trying to make it possible to see what build we had currently deployed.

I can either (a) not push this out (b) change the url (c) update the test.

What makes sense to you?

akainic:
Closing as this has had no activity for awhile.

PR #189, opened 2016-06-28
adds help message if submission fails

crwallace:
This closes:
- #444: Better error message when submission fails

screen shot 2016-06-28 at 11 27 47 am


akainic:
@crwallace I'm thinking of just adding all of the copy in the blue alert box instead of having some of it be styled as an error message. An error message suggests something the user did wrong and can correct, but in this case it's probably not that and instead something wrong with the system. I think notification or alert box styling is better here for the full message.

crwallace:
screen shot 2016-06-28 at 11 48 49 am

akainic:
Can we keep the blue box instead of red? It's a lot of red and it looks alarming. The submit button turning red is enough, I think, to suggest that something went wrong. Thoughts?

crwallace:
Well I think it needs to be a color other than blue. Throughout the app and vets.gov we use blue boxes and extra information without action or direction attached to it. The purpose of this is to provide feedback with a direction. I don't think the red is alarming but I don think it makes it very clear that the form did not submit and the user will need to take further action.

akainic:
I'm just worried about what happened when we had that one user testing session where the veteran was very dismayed when he saw the submission fail. Red definitely reads as a warning or something negative, so I don't want to exacerbate the frustration a veteran might feel at the form failing to submit. Perhaps @mariana-almeida, @gnakm, or @maryannbrody have thoughts?

crwallace:
I completely understand that. I just don't want to use UI patterns that we are not using for application feedback as application feedback because we don't want a negative response to appear negative. The user not be able to submit is bad and I don't want them to be confused about what has happened and not know next steps because the red is "too much bad".

maryannbrody:
@akainic @crwallace -- What if we redirect to an error page? This is more of a "system" error (like a 404) more so than a form validation error. (Unless the error can be caused by missing something in the form, in which case _ideally_ we'd jump users back up to that field.

crwallace:
@maryannbrody Thats a great idea! And maybe in the next iteration we can build something that will fill out the PDF give you the option to print from that screen. But for now we just display the message.

mariana-almeida:
I agree with @crwallace 's reasoning about how to display the message. I wonder if we want to be more explicit with this text and actually say: "system error". I think we want them to know the problem was on our side. The problem didn't stem from their data, or their internet connection or whatever.

akainic:
@mariana-almeida @crwallace @maryannbrody Okay, sounds like keeping the styling as is for now, possibly creating a new page to redirect to in the future, but I don't think we have time to complete that today. Will add a ticket to revisit post-launch. Should we add some words about it being a system error? Perhaps `We regret we weren't able to process your application at this time due to a system error`?

mariana-almeida:
"Due to a system error, we weren't able to process your application. Please try again later. 

We apologize for the inconvenience. If you'd like to complete this form by phone, please call 877-222-VETS (8387), and press 2, from 8:00 a.m. to 5:00 p.m. (EST)."

You may omit ", from 8:00 a.m. to 5:00 p.m. (EST)" if we can't add times they are open. @akainic is checking on that.

Rationale of the first sentence - since users scan when they read, you want to put the most important parts first  "Due to system error" is more prominent in the beginning of the sentence. Also "try again later" is more prominent at the end of the sentence. 

I checked the style guide for phone number and hours style, also we're using "press 2", not "option 2." Thanks for looping me in!! Is there anymore you'd like me to review?

akainic:
@koconnor719 is finding out about what hours the call center is open. While we wait on the answer to that, let's remove the hours, since that will match what we have on the introduction page. And then we can update with the hours once we have a definitive answer on that.

crwallace:
Current message:
screen shot 2016-06-28 at 4 58 13 pm

src/client/components/HealthCareApp.jsx
```diff
@@ -187,6 +188,10 @@ class HealthCareApp extends React.Component {
             beforeText="&#10003;"/>
       );
     } else {
+      submitMessage = (<div className="usa-alert usa-alert-error">
+        <p><strong>Due to a system error, we weren't able to process your application. Please try again later.</strong></p>
+        <p>We apologize for the inconvenience. If you'd like to complete this form by phone, please call 877-222-VETS (8387) option 2.</p>
```
akainic: I think @mariana-almeida said we're using `press 2` not `option 2`.

akainic:
One last comment, but otherwise looks great!

src/client/components/HealthCareApp.jsx
crwallace: Every other place we say "option 2". Should we change all of them?

akainic:
@crwallace Also confirmed hours from Sheila: M-F 7am-7pm CST, Sat 9-530p CST

mariana-almeida:
I just got the style guide version of the hours, per Beth Potts / Chicago Manual of Style.

Space allowing it would go like this: 

Monday - Friday, 7:00 a.m. - 7:00 p.m. (CST), and Saturday, 9:00 a.m. - 5:30 p.m. (CST)

PR #193, opened 2016-06-29
Inconsistence phrasing, correcting to be consistent

akainic:


akainic:
TBR later, only small wording changes.

PR #196, opened 2016-06-29
Fix SpouseFinancial Details exposed by testing

markolson:
- IncomeType should apparently be 7 instead of 12 for everything.
- Spouse should have a relationship of type 2


src/server/enrollment-system.js
```diff
@@ -365,7 +366,7 @@ function veteranToSpouseFinancials(veteran) {
     grossIncome: veteran.spouseGrossIncome,
     netIncome: veteran.spouseNetIncome,
     otherIncome: veteran.spouseOtherIncome
-  });
+  }, { grossIncomeType: 7 });
```
markolson: ignore this - leftover from when I thought we had to override the types based on dependent/spouse. Will push up a new diff shortly.

src/server/enrollment-system.js
```diff
@@ -192,6 +192,7 @@ function veteranToSpouseInfo(veteran) {
       middleName: veteran.spouseFullName.middle,
       familyName: veteran.spouseFullName.last,
       suffix: veteran.spouseFullName.suffix,
+      relationship: 2,
```
akainic: Just want to verify with Mehedi whether or not we should always pass 2 or if it should be based on the relationship status.

src/server/enrollment-system.js
```diff
@@ -222,7 +223,7 @@ function resourceToIncomeCollection(resource) {
   if (resource.grossIncome > 0) {
     incomeCollection.push({
       amount: resource.grossIncome,
-      type: 12, // Total Employment Income TODO is this right?
+      type: 7, // Total Employment Income TODO is this right?
```
akainic: Can we remove the `TODO` since we know this is right now?

robbiethegeek:
The code looks fine, what about the other data point you mentioned in the slack channel? 

that ​_does not_​ include the contributedToSpouse=false change that was also done in the fixed submission I sent over

[4:01]  
since that was `false` in the VOA submission

markolson:
@robbiethegeek there were 3 main differences we identified between our submission and the VOA submission: these two, and a difference where we're submitting <contributedToSpousalSupport>false</false> and VOA is adding `<contributedToSpouse>false</contributedToSpouse>`. Because they both have `false` values I'm assuming that it's less important than the missing `relationship` field and mis-categorized income type right now. I also need to look more closely at the specification to see if there's a difference between the meaning of the two fields, and if they're required under different situations and it felt a bit out-of-scope for this PR.

akainic:
@markolson We should definitely email Mehedi with that question. Do you think it's worth still including in this PR in case it does have something to do with the spouse income not being submitted until we hear back from Mehedi?

markolson:
I think that we're ok merging, but will defer if anyone thinks otherwise. The field being optional and both having `false` values inclines me to believe that they aren't important.

akainic:
LGTM!

PR #197, opened 2016-06-29
Move GA tracking code for successful submissions inside response

akainic:
GA was not tracking the submission id value. I think this is because the `dataLayer.push` was running before the response was received, therefore it didn't have the value of the id yet. Moving the `dataLayer.push` inside the response and sending the id directly from the response data to GA instead of using `this.props.uiState.submission.id`.


robbiethegeek:
LGTM

robbiethegeek:
let's :shipit: merge when tests pass

PR #198, opened 2016-06-29
Adds Maintenance page code

crwallace:
screen shot 2016-06-29 at 3 55 50 pm


joduinn:
Possible tweak to change "You may call" to "In the meanwhile, you can still call ..."  ?

Only a nit, so feel free to ignore. And yes, I like it!

staceylanger:
Agree with John's changes. I would simplify by removing "Bear with us".

crwallace:
Copy edits below:
screen shot 2016-06-29 at 5 22 15 pm

public/index.html
```diff
@@ -139,7 +141,34 @@ <h2 class="sr-only">Apply for Health Care</h2>
...
+  </div>
+
+<div class="main home" role="main">
+
+  <div class="section main-menu">
+    <div class="row">
+      <div class="small-12 columns">
+        <div style="padding: 2em 0;">
+        <h3>We're sorry. The healthcare application is currently down while we fix a few things. We will be up and running as soon as we can.</h3>
+        <h4>In the meantime, you can still call 1-877-222 VETS(8387) press 2 to complete this application over the phone.</h4>
```
akainic: Something looks odd with the telephone number. Maybe we need a `-` between `222` and `VETS` and a space between `VETS` and `(8387)`. Also it's `and press 2` afterwards, missing an `and`.

public/index.html
```diff
@@ -139,7 +141,34 @@ <h2 class="sr-only">Apply for Health Care</h2>
     </div>
   </div>
 </div>
+<!-- HCA Application End -->
 
+<!-- Maintenance Page Start -->
+  <!--<div class="splash">
```
akainic: We're keeping this commented out so we can uncomment it and deploy if anything goes wrong?

public/index.html
```diff
@@ -139,7 +141,34 @@ <h2 class="sr-only">Apply for Health Care</h2>
...
+    </div>
+  </div>
+
+<div class="main home" role="main">
+
+  <div class="section main-menu">
+    <div class="row">
+      <div class="small-12 columns">
+        <div style="padding: 2em 0;">
+        <h3>We're sorry. The healthcare application is currently down while we fix a few things. We will be up and running as soon as we can.</h3>
```
akainic: Maybe `We will be back up as soon as we can` instead of `up and running`? More clear?

public/index.html
```diff
@@ -139,7 +141,34 @@ <h2 class="sr-only">Apply for Health Care</h2>
...
   </div>
 </div>
+<!-- HCA Application End -->
 
+<!-- Maintenance Page Start -->
+  <!--<div class="splash">
+    <div class="row">
+      <div class="small-12 columns">
+        <div class="pitch">
+          <h2 class="tagline"><span>Wednesday, June 29th, 2016 at 3:38 PM EST</span></h2>
```
akainic: Can we enter the current time somehow instead of hard-coding this?

public/index.html
crwallace: Correct!

public/index.html
crwallace: @maryannbrody weigh in if you have feelings about this.

public/index.html
crwallace: We would need to capture the time of the deploy. We are trying to give the user an idea of how long that message has been up. Do we have a place where that timestamp is getting stored?

crwallace: @maryannbrody -- Do we think it is crucial for the user to know the time? I am thinking about removing it from the header. I started looking through other examples of system maintenance messages and very few have a timestamp.

public/index.html
maryannbrody: @crwallace - no strong preference. @akainic's suggestion works for me.

crwallace:
This closes: 
- 435: Create a redirect page if app is down

akainic:
Looks great! Updating branch and then will merge once tests pass.

PR #200, opened 2016-07-01
Automatically retry travis builds on sauce connect timeout

alexose:
I think this fixes https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/461, but it's pretty tough to know for sure unless Saucelabs continues to have timeout issues.

Approving this PR is a vote for "travis_retry is an acceptable thing to use here".


markolson:
LGTM

PR #201, opened 2016-07-05
HCA UI Fixes

crwallace:
This closes:
- 459: Disable button if a user gets "Send Failed"
- 437: Hide spouse support if you lived with spouse
- 463: Add OMB control number

### DIFF
--- a/assets/css/hca.scss
+++ b/assets/css/hca.scss
@@ -108,6 +108,10 @@ button.hca-button-green {
   background: $color-green;
 }
 
+button.hca-button-disabled {
+  pointer-events: none;
+}
+
 .progress-box {
   padding: 1rem 2rem;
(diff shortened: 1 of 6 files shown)


akainic:
LGTM! The only question I have is about the placement of the OMB number. Does it need to go in the header for that page? Seems like it might be better at the bottom of the introduction text or right below the header, something not so prominent.

crwallace:
I thought in the header made the most sense since it is an OMB number for the form. They are realted and any other place I put it felt random and without context.

akainic:
Are OMB numbers used by veterans or other people that will use this form to identify the form? Does it mean something to those people? I'm not totally clear what it is used for, other than it's relation to PRA.

akainic:
Let's merge as is and we can revisit the question about OMB numbers.

PR #204, opened 2016-07-07
Exclude all spouseFinancialsList if unmarried

markolson:
`spouseFinancialsList` was being included even if the `maritalStatus` was set to something other than `Married` or `Separated`, where financials/personal data may be provided. The logic around Spouse data now more closely aligns with how Children are serialized.


test/data/conformance/no-spouse.json
```diff
@@ -0,0 +1,116 @@
...
+    "city": "Washington",
+    "country": "USA",
+    "state": "DC",
+    "zipcode": "20005"
+  },
+  "email": "foo@example.com",
+  "homePhone": "1231241234",
+  "mobilePhone": "1235551234",
+  "understandsFinancialDisclosure": true,
+  "dateOfMarriage": "1983-05-10",
```
akainic: Should this be included if there is no spouse?

test/data/conformance/only-vet.json
```diff
@@ -0,0 +1,75 @@
...
+    "city": "Washington",
+    "country": "USA",
+    "state": "DC",
+    "zipcode": "20005"
+  },
+  "email": "foo@example.com",
+  "homePhone": "1231241234",
+  "mobilePhone": "1235551234",
+  "understandsFinancialDisclosure": true,
+  "dateOfMarriage": "1983-05-10",
```
akainic: Same here?

test/data/conformance/only-vet.json
```diff
@@ -0,0 +1,75 @@
...
+  "sameAddress": true,
+  "cohabitedLastYear": true,
+  "provideSupportLastYear": false,
```
akainic: Or these?

src/server/enrollment-system.js
```diff
@@ -362,6 +359,10 @@ function veteranToDependentFinancialsCollection(veteran) {
  * @returns {Object} ES system dependentFinancialsCollection message
  */
 function veteranToSpouseFinancials(veteran) {
+  if (!_.includes(['Married', 'Separated'], veteran.maritalStatus)) {
```
markolson: `spouseFinancials` is the container for all spouse-related data, including name, address, city of birth, etc.. Returning `undefined` for anything other than Married/Separated lines up with the checks on the React & frontend validation side [like in this examples](https://github.com/department-of-veterans-affairs/healthcare-application/search?utf8=✓&q=%22this.props.data.maritalStatus.value+%3D%3D%3D+%27Married%27%22)

src/server/enrollment-system.js
```diff
@@ -185,31 +185,28 @@ function yesNoToESBoolean(yesNo) {
  * @returns {Object} ES system spouseInfo message
  */
 function veteranToSpouseInfo(veteran) {
-  if (_.includes(['Married', 'Separated'], veteran.maritalStatus)) {
```
akainic: Why did we remove this check?

src/server/enrollment-system.js
akainic: Looks like if they aren't married, this will cause none of the financial info to be returned, since it looks like on lines 925-926 below that some children financial info is returned. Do we still want to return that regardless of whether or not they have a spouse?

test/data/conformance/only-vet.json
markolson: Even though they don't impact the output (which is conditional on other items), I really should remove these and clean these samples up. Will do that now

src/server/enrollment-system.js
markolson: Dependent financial info will still be returned if they're unmarried, as that comes from the `veteranToDependentFinancialsCollection` function, and the `child-financial` examples still covers that case. This change should continue to set `hasIncomeData` correctly if no income is present, but also not include the `spouseFinancialsList` at all if there is no spouse.

src/server/enrollment-system.js
markolson: Moved the negated version up a level into `veteranToSpouseFinancials`

src/server/enrollment-system.js
akainic: Gotcha. Okay now I'm seeing that `dependentFinancials` and `numberOfDependentChildren` are being included elsewhere. Was just a little confused to see those returned with spouseFinancials.

akainic:
Looks good to me! Let's get this merged and then test on staging to make sure it works before deploying to prod.

PR #206, opened 2016-07-07
Update production 070716

akainic:


markolson:
Diff looks correct to me

PR #207, opened 2016-07-07
Require acknowledgement of financial disclosure

austin-martinez:
Completes #404


src/client/components/household-information/FinancialDisclosureSection.jsx
```diff
@@ -70,12 +69,11 @@ class FinancialDisclosureSection extends React.Component {
           </div>
 
           <div className="input-section">
-            <ErrorableRadioButtons required
-                errorMessage={validateIfDirty(this.props.data.understandsFinancialDisclosure, isNotBlank) ? '' : 'Please select a response'}
+            <ErrorableCheckbox required
+                errorMessage={isValidFinancialDisclosure(this.props.data) ? '' : 'Please select a response'}
```
markolson: Should this language change to something like 'Please acknowledge this requirement'?

austin-martinez: Does anyone else have a suggestion? I like the above so will change to that unless hear otherwise

akainic:
@austin-martinez We got new intel on this today from the HEC that might change our solution here. Hold tight on merging this until we get a definitive answer! Sorry for the last minute notification!

akainic:
Now that we decided to always show the financial disclosure questions, even if they end up not entering income and expenses, I think the checkbox solution is what we want going forward! So this PR is no longer on hold, feel free to get a review on this and merge it in :).

src/client/components/household-information/FinancialDisclosureSection.jsx
akainic: Above suggestion is good! Would also love input on this from @maryannbrody and @crwallace

markolson:
LGTM once tests pass

Issue #208, opened 2016-07-07
Improve app configuration


ghost:
Right now, the Ansible scripts that deploy the healthcare app do a regex search/replace in `config.js` to configure the app for each deployment environment, which is rather fragile. Looking around, it looks like the `nconf` package provides a cleaner solution:

http://stackoverflow.com/questions/5869216/how-to-store-node-js-deployment-settings-configuration-files

Would it be possible to integrate this package into the healthcare app? For reference, this is what we're doing right now:

The replacement code lives here:

https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/roles/vetsgov/healthcare-app/tasks/main.yml#L63

And takes values from here:

https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/roles/vetsgov/healthcare-app/defaults/main.yml#L12

At @ShawnArnwine's suggestion, including @alexose.

alexose:
Excellent!  I would love to replace our existing config scripts.

`nconf` seems like a good solution here.  Rather than overwriting config.js, we could pass the relevant values via environment variables or command line arguments.  Potentially sensitive settings would still live in the devops repo.  Is this what you had in mind?

(It would also be easy to accomplish this with native node, but I agree that `nconf` looks cleaner.)

ghost:
Yeah, both environment-specific and sensitive settings would reside in the devops repo (probably with defaults appropriate for running a local dev copy). It would be nice if the `healthcare-application` repo just had a blank `config.json`, and did something like `nconf.file({ file: 'config.json' });`. That way, Ansible could just overwrite `config.json` with the appropriate values.

alexose:
It seems like the problem that the regexes solve is that `devops` doesn't want to duplicate all of the configuration options in `healthcare-application`, but still wants to be able to override them.  So yeah, I'll make the application check for the existence of a `config.json`, which will extend whatever is in `config.js`.

Hopefully that makes sense!  PR incoming...

alexose:
@mechaxl #209 is merged.  Changing over the regexes next.

alexose:
Blocked on getting AWS credentials so I can test ansible.  Hopefully should have that cleared up soon

ShawnArnwine:
What other AWS creds do you need?

alexose:
Just needed an aws key/secret.  @mechaxl hooked me up.

Running into a different problem now:

```
[...]

PR #209, opened 2016-07-08
add the ability to use outside configuration options from a file

alexose:
Addressing #208. This PR adds the ability to pass configuration options in a file, which is useful for the devops folks.

Say you want to override the port:  Just add your configuration options in `./config.extra.js`:

```
module.exports = {
  port : 3001
};
```

Then run the server.  Observe as it is now running on port 3001!


markolson:
LGTM after a rebase/merge

PR #210, opened 2016-07-08
Adds Conditional logic for childCohabitedLastYear in child section

crwallace:
Closes:
Conditional logic needed in child section #376


akainic:
Looks great! Is there one more reference to `client.click('input[name="childCohabitedLastYear-0"]');` in `common.js`? Also, let's wait to merge until tomorrow so as not to introduce anything new with the fix we want to deploy today.

akainic:
LGTM!

PR #214, opened 2016-07-11
Remove the entire Financial block if no financials are provided

markolson:
If no actual income data is provided, remove the entire `financialsInfo` block from the submission.


akainic:
LGTM assuming tests pass!

PR #215, opened 2016-07-11
Include Associations

markolson:
Including associations fails the XML Schema validation test. Setting it as an empty array passes, as does including a single association object instead of an array.

The golden JSON test could be fixed as well, but I'd prefer to not change that until if we know the data is correct or not.


src/server/enrollment-system.js
```diff
@@ -1062,8 +1086,17 @@ function veteranToFinancialsInfo(veteran) {
...
+function veteranToAssociationCollection(veteran) {
+  const associations = [];
+  const children = _.compact(veteran.children.map((child) => { return childToAssociation(child); }));
+  const spouse = spouseToAssociation(veteran);
+  associations.concat(children);
+  if (spouse) { associations.push(spouse); }
+  if (_.isEmpty(associations)) {
+    return undefined;
+  }
+  const result = associations.length === 1 ? associations.pop() : associations;
```
robbiethegeek: @akainic like my use of a ternary :)

src/server/enrollment-system.js
```diff
@@ -1094,13 +1092,11 @@ function veteranToAssociationCollection(veteran) {
...
   if (spouse) { associations.push(spouse); }
-
   if (_.isEmpty(associations)) {
     return undefined;
   }
-  const result = associations.map((association) => { return { association }; });
-  console.log(result);
-  return result;
+  const result = associations.length === 1 ? associations.pop() : associations;
+  return { association: result };
```
markolson: I think the SOAP library will handle this correctly if it's left as a single item array, but I'll write some tests around it to verify

markolson: @robbiethegeek after updating the conformance test XML, I changed this to just include `associations` as-is, and the results came out the same for `no-children.xml` which only has one `association` record.

robbiethegeek: 👍 cool wasn't sure if we needed to manually do it or the soap library would thanks for checking.

robbiethegeek:
LGTM @akainic want to give it a quick once over?

akainic:
LGTM! Let's merge and test in staging and verify with Josh.

PR #216, opened 2016-07-11
Update production 071116

akainic:


markolson:
Financial changes LGTM

PR #217, opened 2016-07-11
Update index.html

ayaleloehr:
Removing TM from vets.gov footer, as we only need the trademark symbol in one place on the site and it is already located at https://www.vets.gov/employment/

Agreed to by the content team in a parallel change here: department-of-veterans-affairs/vets-website#2551


akainic:
LGTM!

PR #218, opened 2016-07-11
adds Veterans Crisis Line to HCA

crwallace:
This adds VCL to HCA footer.

screen shot 2016-07-11 at 5 31 39 pm

@webinista-- Would you mind taking a first pass at this to make sure that I implemented everything the same at vets-website?


webinista:
Gave a quick look-through. It looks good to me.

PR #226, opened 2016-07-19
Force all names to UPPERCASE before transmission

markolson:


src/server/utils/validations.js
```diff
@@ -49,6 +49,14 @@ const validations = {
     }
     return validatedString;
   },
+
+  validateName: (data, count, nullable) => {
```
austin-martinez: is there another file/name that makes sense for this function? it isn't really validating the name as much as it is formatting it. maybe just add it to the enrollment system file as a function called `formatName` or similar? also, you aren't passing in the last 2 arguments to this function, so you could always pass in `nullable=false` to the `validateString` method, which will then always return a string and the `isEmpty` check could be removed from this function

akainic: Maybe calling it `nameToUpperCase`, just to be super explicit about what this function does?

markolson: `nameToUpperCase` wouldn't reflect what it's doing though, as it's still passing the data through to do the length and nullability checks. This function does also receive the 2nd and 3rd params (see line 430, for instance). I could consolidate the functions, but I think there's a use in keeping a separate `validateString` function.

akainic:
LGTM! Let's update the merge conflicts and get this merged!

PR #227, opened 2016-07-19
HCA content edits

crwallace:


crwallace:
@staceylanger-- Can you take a quick peek at these changes to make sure that we don't need to run anything by counsel? I think most of it is minor but just wanted your eyes on this.

staceylanger:
Thanks, @crwallace. This language came directly from legal so I would want to check with her if we change it:
 Federal law provides criminal penalties, including a fine and/or imprisonment for up to 5 years, for concealing a material fact or making a materially false statement. 

So I can either have her review the new language, or you can make all the changes except the changes to that line and move forward.

src/client/components/household-information/SpouseInformationSection.jsx
```diff
@@ -197,11 +197,11 @@ class SpouseInformationSection extends React.Component {
       content = (<fieldset>
         <legend>Spouse's Information</legend>
         <p>(<span className="hca-required-span">*</span>) Indicates a required field</p>
-        <p>Please fill this out to the best of your knowledge. The more accurate your responses, the faster your application can proceed.</p>
+        <p>Please fill this out to the best of your knowledge. The more accurate your responses, the faster we can process your application.</p>
         <div className="input-section">
           <ErrorableSelect
               errorMessage={validateIfDirty(this.props.data.maritalStatus, isNotBlank) ? undefined : 'Please select a marital status'}
-              label="Current Marital Status"
+              label="Current marital mtatus"
```
akainic: Typo: should be `status`

src/client/components/insurance-information/AdditionalInformationSection.jsx
```diff
@@ -65,7 +65,7 @@ class AdditionalInformationSection extends React.Component {
               value={this.props.data.vaMedicalFacility}
               facilityState={this.props.data.facilityState}
               onValueChange={(update) => {this.props.onStateChange('vaMedicalFacility', update);}}/>
-          OR <a target="_blank" href="/facility-locator">Go to the VA Facility Locator</a>
+          OR <a target="_blank" href="http://www.va.gov/directory/guide/home.asp">Find locations with the VA Facility Locator</a>
```
akainic: Hm do we need to use the va.gov facility locator if we have one of our own? I think we should use our facility locator tool going forward, unless there is strong reason not to?

akainic:
@crwallace Couple of nits. If @staceylanger is good with this language it looks good to me otherwise!

akainic:
Rolled back that one content change that Stacey identified to be what it originally was, so we don't need any new review from legal on that.

PR #228, opened 2016-07-20
Use dirty flag on understandFinancialDisclosure checkbox

austin-martinez:
Use dirty flag on understandFinancialDisclosure checkbox to prevent error message on first load

### DIFF
--- a/src/client/components/household-information/FinancialDisclosureSection.jsx
+++ b/src/client/components/household-information/FinancialDisclosureSection.jsx
@@ -13,7 +13,7 @@ import { veteranUpdateField } from '../../actions';
  */
 class FinancialDisclosureSection extends React.Component {
   errorMessage(data) {
-    const field = this.props.data.understandsFinancialDisclosure;
+    const field = data.understandsFinancialDisclosure;
     let error;
 
     // check if true only if dirty (prevents error message on first load of section)
--- a/src/client/components/household-information/FinancialDisclosureSection.jsx
+++ b/src/client/components/household-information/FinancialDisclosureSection.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import _ from 'lodash';
 
 import ErrorableCheckbox from '../form-elements/ErrorableCheckbox';
 import { isValidFinancialDisclosure } from '../../utils/validations';
--- a/src/client/components/household-information/FinancialDisclosureSection.jsx
+++ b/src/client/components/household-information/FinancialDisclosureSection.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import _ from 'lodash';
 
 import ErrorableCheckbox from '../form-elements/ErrorableCheckbox';
-import { isValidFinancialDisclosure } from '../../utils/validations';
+import { validateIfDirty } from '../../utils/validations';
 import { veteranUpdateField } from '../../actions';
 
 /**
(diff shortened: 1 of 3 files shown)


austin-martinez:
@akainic i wasn't able to test with pre-prod locally due to not having VA network access, but the JSON payload looks good and this value isn't sent to ES so should be no issues there

src/common/veteran.js
```diff
@@ -259,7 +259,7 @@ const completeVeteran = {
     value: '1235551234',
     dirty: false
   },
-  understandsFinancialDisclosure: true,
+  understandsFinancialDisclosure: makeField(true),
```
akainic: I think this should instead be: 
```javascript
understandsFinancialDisclosure: {
  value: false,
  dirty: false
},
```

akainic:
I think there is a simpler way of doing this. I think you don't need to add the `errorMessage` function in `FinancialDisclosureSection.jsx`. Instead, if you look in `src/client/utils/validations.js`, you'll see a function called `validateIfDirty`. We can use that when validating if the Financial Dislcosure Section is valid (the function in `validations.js` called `isValidFinancialDisclosure`. 

What I did was create a new function in `validations.js` called `isNotFalse`:

``` javascript
function isNotFalse(value) {
  return value !== false;
}
```

^^^ Be sure to export this function at the bottom of the file.

Then, `isValidFinancialDisclosure` can become:

``` javascript
function isValidFinancialDisclosure(data) {
  return validateIfDirty(data.understandsFinancialDisclosure, isNotFalse);
}
```

And that should do it! All the other changes you made looked good, I just don't think you need to create that other function and check if the field is dirty in the `FinancialDisclosureSection.jsx`. Let me know if that doesn't make sense.

src/common/veteran.js
austin-martinez: `makeField` sets the dirty flag to false by default. i kept the value as true because that is what it was set as before, is there a reason to set it to false? this object represents a completed veteran form so i think `value: true` makes sense here, with `value: false` for the blank veteran object as set on L56

akainic: Ah yes, it should be `true`, you're right on that. But let's not use `makeField` here and instead enter the JSON object directly.

src/client/components/household-information/FinancialDisclosureSection.jsx
```diff
@@ -73,8 +74,8 @@ class FinancialDisclosureSection extends React.Component {
                 errorMessage={isValidFinancialDisclosure(this.props.data) ? '' : 'Please acknowledge this requirement'}
                 label="I understand VA is not currently enrolling new applicants who decline to provide their financial information unless they have other qualifying eligibility factors."
                 name="understandsFinancialDisclosure"
-                checked={this.props.data.understandsFinancialDisclosure}
-                onValueChange={(update) => {this.props.onStateChange('understandsFinancialDisclosure', update);}}/>
+                checked={this.props.data.understandsFinancialDisclosure.value}
+                onValueChange={(update) => {this.props.onStateChange('understandsFinancialDisclosure', { value: update, dirty: false });}}/>
```
akainic: Do you need to pass `{ value: update, dirty: false }`? I think just `update` will work.

austin-martinez: the checkbox only sends the value and not the field with the dirty flag (this is different from the other field inputs), so with just update we would lose the dirty flag on the field

akainic: ah gotcha, thanks!

akainic:
LGTM! Once tests pass feel free to merge

PR #229, opened 2016-07-21
Add contactType to associations

markolson:


akainic:
LGTM! Let's merge once tests pass and request a deploy to staging so we can test

PR #231, opened 2016-07-25
Fix logic around State/Province Zip/Postal requirements

markolson:
Once this is merged in, the Address form sections for both the
vet and their spouse will have differing requirements based on
if the country selected in USA. If so, State and ZIP will be
required. If USA, Canada or Mexico are selected, State/Province
will be a dropdown of permitted values, otherwise it will become
a free text entry field, and ZIP/Postal will be optional.

Addresses are also included in Spouse Associations now.


src/client/components/questions/Address.jsx
```diff
@@ -83,23 +94,56 @@ class Address extends React.Component {
...
+    if (this.state.requireStateZip) {
+      secondaryForm = (
+        <div>
+          <ErrorableSelect errorMessage={this.validateAddressField(this.props.value.state) ? undefined : 'Please enter a valid state'}
+              label="State/Province"
+              name="state"
+              autocomplete="address-level1"
+              options={stateList}
+              value={this.props.value.state}
+              required={this.state.requireStateZip ? this.props.required : false}
```
austin-martinez: this conditional will always be true as this is in a block dependent on the condition being true L101

src/client/components/questions/Address.jsx
```diff
@@ -83,23 +94,56 @@ class Address extends React.Component {
...
 
-        <ErrorableTextInput errorMessage={this.validateAddressField(this.props.value.zipcode) ? undefined : 'Please enter a valid ZIP code'}
-            label="ZIP Code"
-            name="zip"
-            autocomplete="postal-code"
-            field={this.props.value.zipcode}
-            required={this.props.required}
-            onValueChange={(update) => {this.handleChange('zipcode', update);}}/>
+          <ErrorableTextInput errorMessage={this.validateAddressField(this.props.value.zipcode) ? undefined : 'Please enter a valid ZIP code'}
+              label={this.state.requireStateZip ? 'ZIP Code' : 'Postal Code'}
```
austin-martinez: same as above, L118 also checks this condition

akainic:
FYI just got off a call and Josh mentioned that postal code is a required field is it is a foreign country. I think we originally thought province and postal code are optional if a foreign country, but only province is optional, postal code is required.

src/client/components/questions/Address.jsx
markolson: You're right - that's leftover when I was trying to figure out why validations were still failing.

markolson:
@austin-martinez fixed the things you found

src/client/components/questions/Address.jsx
```diff
@@ -25,12 +28,20 @@ class Address extends React.Component {
   // TODO: Look into if this is the best way to update address,
   // it is incredibly slow right now
   handleChange(path, update) {
+    // if the country is US, Canada, or Mexico, then the
+    // state/provice + postal is required
+    if (path === 'country') {
+      this.setState({ requireStateZip: _.hasIn(states, update.value) });
```
akainic: I would prefer not to store this in state if possible. Can we instead just check the value of `this.props.value.country` directly?

markolson: Sure thing

markolson:
While testing, I found an issue with form validation I'll fix in the AM

src/client/utils/validations.js
```diff
@@ -106,8 +106,12 @@ function isValidEmail(value) {
 // TODO:  1. what is a valid address?
 //        2. 6 arguments to a function is ugly...
 //        3. argument order is now based on form order... using
-function isValidAddress(street, city, country, state, zipcode) {
-  return isNotBlank(street) && isNotBlank(city) && isNotBlank(country) && isNotBlank(state) && isNotBlank(zipcode);
+function isValidAddress(street, city, country, state, zipcode, postalCode) {
+  const initialOk = isNotBlank(street) && isNotBlank(city) && isNotBlank(country);
+  if (country === 'USA') {
+    return initialOk && isNotBlank(state) && isNotBlank(zipcode);
```
akainic: Should we also add a validation on ZIP code here?

markolson:
@akainic tests passed 🎉

src/client/utils/validations.js
```diff
@@ -468,7 +475,6 @@ export {
   isValidMonetaryValue,
   isValidPhone,
   isValidEmail,
-  isValidAddress,
```
akainic: Do we need to export `isValidAddressField` here?

test/e2e/utils/common.js
```diff
@@ -170,6 +170,12 @@ const testValues = {
   campLejeune: false
 };
 
+function selectDropdown(client, field, value) {
+  client.click(`select[name='${field}']`)
+    .click(`select option[value='${value}']`)
+    .keys(['\uE006']);
```
akainic: Is this to fix the Selenium errors you were seeing?

src/client/utils/validations.js
markolson: It's only used internally to check Vet/Spouse addresses

test/e2e/utils/common.js
markolson: Yes. Firefox has an issue with dropdown selections triggering events when being driven by selenium, and this is a bit of a workaround I found.

src/client/utils/validations.js
akainic: Oh. Just realized `isValidAddressField` here isn't even being used at all? There's a different function defined in `Address` called `validateAddressField` that's actually being used.

markolson: `isValidAddressField` is called via `isValidForm`, which is called when the 'Continue' button is clicked

akainic: Ah right, thanks!

akainic:
LGTM!

PR #232, opened 2016-07-26
PrettyPrint XML conformance files

markolson:
Going format, having the XML files pretty printed will help with readability for the diffs, letting us actually see what changes are happening in the serialization process when reviewing changes here


akainic:
LGTM!

PR #233, opened 2016-07-27
Change JSON validation to account for postalCode/provinceCode

markolson:
Submissions were being rejected when zip code was missing

### DIFF
--- a/src/common/schema/application.js
+++ b/src/common/schema/application.js
@@ -13,6 +13,11 @@ const countryStateProperites = _.map(options.states, (value, key) => ({
     state: {
       'enum': value.map(x => x.value)
     },
+    zipcode: {
+      type: 'string',
+      minLength: 1,
+      maxLength: 50
+    }
   }
 }));
 countryStateProperites.push(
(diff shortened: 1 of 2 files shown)


src/common/schema/application.js
```diff
@@ -47,19 +57,12 @@ module.exports = {
           type: 'string',
           minLength: 1,
           maxLength: 51
-        },
```
akainic: Why are only street and city listed in the properties here?

akainic: Oh because they're the only ones that are consistent across the 2 options for country fields?

markolson: Yep! The frontend handles the validations _enough_ that I'm happy changing this up, especially since it doesn't give a useful error message on failure.

akainic: Makes sense, sorry I was slow in understanding what was going on here :)

akainic:
LGTM! Feel free to merge once tests pass

test/server/enrollment-system.spec.js
```diff
@@ -45,6 +45,9 @@ describe('enrollment-system base tests', () => {
     for (const filename of checks) {
       it(`should serialize ${filename} correctly`, (done) => {
         const application = require(`../data/conformance/${filename}`);
+        const valid = validate(application);
+        chai.assert.isTrue(valid, JSON.stringify([validate.errors, application], null, 2));
```
markolson: This was failing on australian-vet before the json-schema changes.

PR #237, opened 2016-08-01
Exclude SpouseFinancials if there's no SpouseFinancial data

markolson:
Previously we were excluding the SpouseFinancial section iff the entire FinancialInfo section was excluded - this makes it a bit more fine-grained.


akainic:
Just wondering if we want to do the same thing for dependents? Otherwise LGTM!

markolson:
Good catch - I'll write up a test (and possible fix) for that and add it to this PR

markolson:
@akainic this is ready for another look when you have a minute to check over the diffs.

akainic:
LGTM!

PR #240, opened 2016-08-02
rate limit healthcheck/es calls

alexose:
This prevents us from accidentally DDOSing `healthcheck/es` if we start hitting the endpoint a lot.

Related to https://github.com/department-of-veterans-affairs/devops/issues/307

### DIFF
--- a/config.js
+++ b/config.js
@@ -77,10 +77,12 @@ const config = {
   },
 
   environment: process.env.NODE_ENV || 'development',
-  // This is a valid form submission ID from the ES system that will be used by the health check to 
+  // This is a valid form submission ID from the ES system that will be used by the health check to
   // verify that the ES system is up.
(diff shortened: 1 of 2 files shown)


akainic:
LGTM, but I would prefer if @robbiethegeek could also take a look at this.

akainic:
@alexose how are we currently using the status endpoint? Is this what you configured in AWS?

alexose:
It's not used by anything at the moment.  Ideally, we'd use it in two places:
- The frontend should check it on page load in order to display a message if ES is down
- AWS should check it every so often to maintain a record of ES uptime

akainic:
Gotcha. This seems like a valuable thing to have in place anyway. We probably want to configure those other uses at some point soon as well. LGTM!

PR #245, opened 2016-08-12
Re-insert Spouse and dependent data into the financial block

markolson:
Per Josh, we should include this data again even if there's no financial data present for the spouse or dependents themselves.

### DIFF
--- a/src/server/enrollment-system.js
+++ b/src/server/enrollment-system.js
@@ -341,9 +341,6 @@ function childToDependentInfo(child) {
  */
 function childToDependentFinancialsInfo(child) {
   const incomes = resourceToIncomeCollection(child);
-  if (!incomes) {
-    return undefined;
-  }
 
   return {
     incomes,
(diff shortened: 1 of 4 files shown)


src/server/enrollment-system.js
```diff
@@ -341,9 +341,6 @@ function childToDependentInfo(child) {
  */
 function childToDependentFinancialsInfo(child) {
   const incomes = resourceToIncomeCollection(child);
-  if (!incomes) {
```
akainic: I'm wondering if this is sufficient, or if we need to include some logic in `veteranToFinancialsInfo` that only includes spouse and dependent financial info if veteran income info also exists? It sounded like from Josh that we should only include spouse and dependent info if they provide veteran income information, but the logic in `veteranToFinancialsInfo` looks for any of them being present? Will this mean that if there is no veteran income, but they do enter spouse information, that it will be included? I don't think it should be in that case.

```
const hasIncomeData = expenses || incomes || hasSpouseIncome || hasDependentFinancials;

if (!hasIncomeData) {
  return undefined;
}
```

markolson: Hm, I do wonder if it's entirely dependent on the vet's financials specifically, or if it's conditional on _any_ financials being present. Would you be ok merging this in and following up/refining with him so they can get some progress on testing?

akainic: Yep, sounds good. I would just ask them to test cases where:
1. No veteran income info is provided and spouse/dependent identity is, but no spouse/dependent income info
2. Veteran income info is provided and spouse/dependent identity is, but no spouse/dependent income info

akainic:
Feel free to merge @markolson so we can deploy to staging and start testing there!

PR #250, opened 2016-08-22
More robust parsing of zip codes

jbalboni:
Addresses this [issue](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/477). 

It's not clear to me if/how we should handle zip codes that aren't 5 or 9 digits. This change should work the same as what would happen with 12345-12 before.

Also, only the first conformance test was actually running, so that's fixed now, too.

### DIFF
--- a/src/server/enrollment-system.js
+++ b/src/server/enrollment-system.js
@@ -52,9 +52,9 @@ function formatAddress(address) {
   };
   if (address.country === 'USA') {
     formatted.state = address.state;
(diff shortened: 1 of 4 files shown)


jbalboni:
@markolson Want to review?

markolson:
I'm not sure we have a great solution to how to handle codes that aren't 5 or 9, when they're in the US. We could probably add a frontend validation for it now that now that we have stronger requirements for US addresses? For now this LGTM though.

PR #252, opened 2016-09-14
Log response body on errors

markolson:
We currently don't get a lot of insight into what's wrong when we get an error back. Logging the body will hopefully give us some actionable details.


markolson:
@jbalboni can you 👀 this and make sure it makes sense?

jbalboni:
@markolson LGTM. I haven't used winston before, so I'm assuming it treats the response reasonably when you pass it as metadata (looks like it should from reading the docs).

PR #254, opened 2016-09-19
Use the same email regex for front and back end validations

markolson:
Looking at Cloudwatch, we have a lot of form validation errors where the emails aren't accepted by what's defined in the json schema. Using the same validation in as the react component should reduce those.

### DIFF
--- a/src/common/schema/application.js
+++ b/src/common/schema/application.js
@@ -243,7 +243,8 @@ module.exports = {
     },
     email: {
       type: 'string',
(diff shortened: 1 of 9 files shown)


jbalboni (approved):
LGTM.

I would put in a plug for testing that something like `foo@example` doesn't pass. I mention it because according to the standard, that's a valid email address. Plus, Chrome's built in email input validation will allow it (unless that changed recently). But we can't do anything with an email that lacks a tld.

markolson:
@jbalboni I added a TODO in my vets-website branch of this to test invalid addresses as well.

PR #257, opened 2016-09-21
Production 09 21 2016

markolson:
- Should fix many of the errors we got on submission because of a regex mismatch between front and back ends.


jbalboni (approved):
LGTM

akainic (approved):
LGTM to me too!

PR #262, opened 2016-10-19
Update facility list

markolson:


akainic (approved):
LGTM! Thanks so much for tackling this! Not sure why that one test is failing though

PR #272, opened 2016-11-03
Add two more address lines

jbalboni:
Closes [491](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/491)

Adds two more lines to the Address component and (I think) makes the backend changes to accommodate that.

screen shot 2016-11-03 at 4 07 00 pm

### DIFF
--- a/test/data/conformance/australian-vet.xml
+++ b/test/data/conformance/australian-vet.xml
@@ -28,6 +28,8 @@
             <eeSummary:city>Sydney</eeSummary:city>
             <eeSummary:country>AUS</eeSummary:country>
             <eeSummary:line1>123 NW 8th St</eeSummary:line1>
+            <eeSummary:line2></eeSummary:line2>
+            <eeSummary:line3></eeSummary:line3>
           </eeSummary:address>
           <eeSummary:givenName>FIRSTSPOUSE</eeSummary:givenName>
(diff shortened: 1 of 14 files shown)


jbalboni:
@akainic @markolson Can you two take a look at this? I'm not super confident in the backend changes, since I don't really know how to test them well.

test/data/conformance/australian-vet.xml
```diff
@@ -28,6 +28,8 @@
             <eeSummary:city>Sydney</eeSummary:city>
             <eeSummary:country>AUS</eeSummary:country>
             <eeSummary:line1>123 NW 8th St</eeSummary:line1>
+            <eeSummary:line2></eeSummary:line2>
```
markolson: Based on how we treat null/non-existant elements elsewhere, I'm thinking that these shouldn't be in the generated XML if they're blank

jbalboni: @markolson Should be fixed now.

markolson (approved):

jbalboni:
@akainic Since @markolson is going to be out, do you want to hold off on merging this? Or should I go ahead when CI is done?

markolson:
@jbalboni merge away, I'd say.

PR #273, opened 2016-11-04
Make expenses required

alexose:
Minor change to the financial disclosure section.  Fixes https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/495

### DIFF
--- a/src/common/schema/application.js
+++ b/src/common/schema/application.js
@@ -395,6 +395,9 @@ module.exports = {
     veteranOtherIncome: ['understandsFinancialDisclosure'],
     spouseGrossIncome: ['understandsFinancialDisclosure'],
     spouseNetIncome: ['understandsFinancialDisclosure'],
-    spouseOtherIncome: ['understandsFinancialDisclosure']
+    spouseOtherIncome: ['understandsFinancialDisclosure'],
+    deductibleMedicalExpenses: ['understandsFinancialDisclosure'],
+    deductibleFuneralExpenses: ['understandsFinancialDisclosure'],
+    deductibleEducationExpenses: ['understandsFinancialDisclosure']
   }
 };
--- a/test/e2e/tests/02-fields-initialized.js
+++ b/test/e2e/tests/02-fields-initialized.js
@@ -248,7 +248,7 @@ module.exports = {
     expectValueToBeBlank(client, 'input[name="deductibleMedicalExpenses"]');
     expectValueToBeBlank(client, 'input[name="deductibleFuneralExpenses"]');
     expectValueToBeBlank(client, 'input[name="deductibleEducationExpenses"]');
-    common.completeDeductableExpenses(client, common.testValues);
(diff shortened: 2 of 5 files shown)


src/common/schema/application.js
```diff
@@ -395,6 +395,9 @@ module.exports = {
     veteranOtherIncome: ['understandsFinancialDisclosure'],
     spouseGrossIncome: ['understandsFinancialDisclosure'],
     spouseNetIncome: ['understandsFinancialDisclosure'],
-    spouseOtherIncome: ['understandsFinancialDisclosure']
+    spouseOtherIncome: ['understandsFinancialDisclosure'],
```
akainic: I think these dependencies are not implemented properly. Looking at the documentation, it looks like the dependency is listed first and the fields they are dependent on are listed in the array: https://spacetelescope.github.io/understanding-json-schema/reference/object.html#property-dependencies.

akainic (commented):
Looks good, one comment!

src/common/schema/application.js
alexose: That was the intent-- `spouseOtherIncome` relies on `understandsFinancialDisclosure`.  However, you're totally right that this doesn't do anything (other than see if understandsFinancialDisclosure exists, which it always does).

The right way to do it would be to actually see if `understandsFinancialDisclosure` is `true` or not, like I'm doing in [this PR](https://github.com/department-of-veterans-affairs/healthcare-application/pull/274).

alexose: Right, I see what you're saying.  I was thinking of it in terms of a user filling out the form, not the final validation.  Will fix in #274!

test/e2e/tests/02-fields-initialized.js
```diff
@@ -248,6 +248,7 @@ module.exports = {
     expectValueToBeBlank(client, 'input[name="deductibleMedicalExpenses"]');
     expectValueToBeBlank(client, 'input[name="deductibleFuneralExpenses"]');
     expectValueToBeBlank(client, 'input[name="deductibleEducationExpenses"]');
+    common.completeDeductableExpenses(client, common.testValues);
```
akainic: nit: typo, it should be `deductible`

test/e2e/utils/common.js
```diff
@@ -392,6 +392,15 @@ function completeChildInformation(client, data, onlyRequiredFields) {
   }
 }
 
+function completeDeductableExpenses(client, data) {
```
akainic: nit: `deductible`

test/e2e/utils/common.js
```diff
@@ -444,6 +453,7 @@ module.exports = {
   completeSpouseInformation,
   completeAnnualIncomeInformation,
   completeChildInformation,
+  completeDeductableExpenses,
```
akainic: nit again

akainic (approved):
Looks good minus a few nits! And the schema conditional logic, but that's to be updated in a separate PR I think.

PR #274, opened 2016-11-04
Use correct conditional logic in jsonschema

alexose:
We might consider adding a test for a `completeVeteran` that fails the conditions as well.


akainic:
@alexose updates on this?

alexose:
Yep, just adding some tests so we know whether or not it's working.  Will update with correct logic in a moment

src/common/schema/application.js
```diff
@@ -390,11 +390,11 @@ module.exports = {
     'dischargeType'
   ],
   dependencies: {
-    veteranGrossIncome: ['understandsFinancialDisclosure'],
-    veteranNetIncome: ['understandsFinancialDisclosure'],
-    veteranOtherIncome: ['understandsFinancialDisclosure'],
-    spouseGrossIncome: ['understandsFinancialDisclosure'],
-    spouseNetIncome: ['understandsFinancialDisclosure'],
-    spouseOtherIncome: ['understandsFinancialDisclosure']
+    veteranGrossIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
```
alexose: Sorry, this is kind of mind-melting...

I think the issue is that we have several fields that depend not only on `understandsFinancialDisclosure`, but also on other factors (for example, `understandsFinancialDisclosure` only needs `spouseGrossIncome` if `maritalStatus !== 'Never Married'`).

Or another way to put it...  `spouseGrossIncome`, if filled out, needs `understandsFinancialDisclosure` to be true, but `understandsFinancialDisclosure` doesn't necessarily need  `spouseGrossIncome` to be filled out.

We could build out the whole dependency tree, but that's not really how we're using JSONSchema... it's mostly there as a quick sanity check before we send the application off to ES.

My recommendation is that we keep going down the path we're already on-- Let's use JSONSchema as a basic rubric and not necessarily try to duplicate all of ES's schema.

akainic: I see what you're saying. I would like @markolson to weight in here when he has a chance.

markolson: That rings true to me. @alexose If you can confirm you tested this against a few docs, 👍

alexose:
This test is still failing on account of voaservice being down... will update when we have a solution

src/common/schema/application.js
```diff
@@ -398,14 +398,14 @@ module.exports = {
...
-    spouseOtherIncome: ['understandsFinancialDisclosure'],
-    deductibleMedicalExpenses: ['understandsFinancialDisclosure'],
-    deductibleFuneralExpenses: ['understandsFinancialDisclosure'],
-    deductibleEducationExpenses: ['understandsFinancialDisclosure']
+    veteranGrossIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
+    veteranNetIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
+    veteranOtherIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
+    spouseGrossIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
+    spouseNetIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
+    spouseOtherIncome: { properties: { understandsFinancialDisclosure: { 'enum': [true] } } },
```
akainic: I think that might fix the schema validation errors we were seeing here, though: https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/501

markolson: @alexose "extra validation on the backend" is fine, but IMO the idea is that this schema is used for making sure frontend tests pass more than submissions hitting on the backend are valid. The backend validation is really (IMO, again) meant to prevent junk spam submissions)

alexose: Yeah, I think we have the same view on this.  We don't really use jsonschema for conformance-- it's mostly there as a sanity check.  This seems like a reasonable amount of sanity to look for.

As to why the `spouse*` fields are behaving differently now, I'm still not sure.  `spouseFullName` isn't a required field, but it's behaving like it is.

leannammiller:
ping for status

markolson:
Going to close this for now because it's not adding quite enough value for the complexity it adds, especially if we also include child financials. We have logic to just ignore these fields if they come in when disclose is false, so, I don't feel like this is a great loss. We can circle back once the vets-website migration is over.

PR #275, opened 2016-11-04
Rename financial field

akainic:
Rename financial field to `discloseFinancialInformation` and change enrollment-system.js to include the financial fields if `discloseFinancialInfo` is true instead of being dependent on whether incomes and expenses exist. I'm not sure if this change makes sense, but since the front-end should be guaranteeting that incomes and expenses exist if they select "yes" for `discloseFinancialInformation`, it should always be the case.

Still working through some conformance test issues, but wanted to push up for review.


akainic:
Tagging @alexose and @markolson (can wait until Monday) for review.

src/server/enrollment-system.js
```diff
@@ -932,21 +932,11 @@ function veteranToFinancialsInfo(veteran) {
...
-  }
   const spouseFinancials = veteranToSpouseFinancials(veteran);
-  const hasSpouseIncome = spouseFinancials && spouseFinancials.spouseFinancials.incomes;
-
-  const hasIncomeData = expenses || incomes || hasSpouseIncome || hasDependentFinancials;
-
-  if (!hasIncomeData) {
-    return undefined;
-  }
+  const discloseFinancials = veteran.discloseFinancialInformation;
```
markolson: Based on my reading of the PR description, I think we want to just do something like
```
if (!veteran.discloseFinancialInformation) { 
    return undefined;
}
```
here so that `marriedLastCalendarYear` and `numberOfDependentChildren` aren't forcing `financialStatement` to be present if `discloseFinancialInformation` is false.

src/server/enrollment-system.js
```diff
@@ -932,21 +932,11 @@ function veteranToFinancialsInfo(veteran) {
...
 
   return {
-    incomeTest: booleanToIncomeTest(hasIncomeData),
+    incomeTest: booleanToIncomeTest(discloseFinancials),
```
markolson: At this point, we should probably change this line to be `incomeTest: { discloseFinancialInformation: true }` because it won't be included either way if `discloseFinancials` is false. Let's just remove the indirection..

test/data/conformance/no-financial.xml
```diff
@@ -96,6 +63,12 @@
           <eeSummary:radiationExposureInd>true</eeSummary:radiationExposureInd>
         </eeSummary:specialFactors>
       </eeSummary:enrollmentDeterminationInfo>
+      <eeSummary:financialsInfo xmlns:eeSummary="http://jaxws.webservices.esr.med.va.gov/schemas">
```
markolson: `discloseFinancialInformation` is false, but `financialsInfo` is still present here. This is the test I sort of want to verify is behaving correctly, as referenced in my other comment.

markolson (changes requested):
I think this is 90% of the way there, but want to verify (and change, if needed) some of the logic. Changes to the test data looks good and thorough!

src/server/enrollment-system.js
akainic: Yep that makes sense!

src/server/enrollment-system.js
akainic: Ah, so you mean if I make the change you identified above, with returning undefined if `!veteran.discloseFinancialInformation`, then we assume it will be true if it reaches this point?

test/data/conformance/no-financial.xml
akainic: Yeah I was very confused about how to fix the conformance tests, this might be included because of the logic changes you identified earlier, so once I make those this should go away, I just copied it in based on wha the failing test expected to see.

src/server/enrollment-system.js
markolson: Yep!

test/data/conformance/no-financial.xml
markolson: Yep, that's exactly how I did it - these conformance tests are totally supposed to "catch" these sort of logic changes, and let you take a step back and look at how different applications change in response to things like the above change in enrollment-systems.

akainic:
@markolson made the changes you requested! Still working on the marital status change (might include that in a separate PR as this one is already large). Let me know how this looks to you so far.

markolson (approved):

markolson:
LGTM! I'll do another review when the marital changes come in

akainic:
@markolson this is ready for your review again!

alexose:
LGTM, again :)

markolson (approved):

PR #276, opened 2016-11-04
Make child income required

alexose:
This addresses [#494](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/494)

There's a [known issue](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/497) that sets each child income field to its error state even before the user has interacted with it.

image

### DIFF
--- a/src/client/components/household-information/ChildIncome.jsx
+++ b/src/client/components/household-information/ChildIncome.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import ErrorableTextInput from '../form-elements/ErrorableTextInput';
 
-import { isBlank, isValidMonetaryValue, validateIfDirty } from '../../utils/validations';
+import { isValidField, isValidMonetaryValue } from '../../utils/validations';
 
 /**
  * Sub-component for children income portion AnnualIncomeSection.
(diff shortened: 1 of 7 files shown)


src/client/components/household-information/ChildIncome.jsx
```diff
@@ -12,26 +12,39 @@ import { isBlank, isValidMonetaryValue, validateIfDirty } from '../../utils/vali
  * `onValueChange` - a function with this prototype: (newValue)
  */
 class ChildIncome extends React.Component {
+
+  getRequiredErrorMessage(field) {
+    let result;
+    if (isValidField(isValidMonetaryValue, field)) {
+      if (field.dirty && field.value === '') {
```
akainic: I feel like we could improve this function if we made us of the validation functions we already have like `validateIfDirty` to keep things consistent, but it's not necessarily a priority.

src/client/utils/validations.js
```diff
@@ -267,10 +267,10 @@ function isValidFinancialDisclosure(data) {
   return validateIfDirty(data.understandsFinancialDisclosure, isNotBlank);
 }
 
-function isValidIncome(income) {
-  return isValidField(isValidMonetaryValue, income.grossIncome) &&
-      isValidField(isValidMonetaryValue, income.netIncome) &&
-      isValidField(isValidMonetaryValue, income.otherIncome);
+function isValidRequiredIncome(income) {
+  return isValidRequiredField(isValidMonetaryValue, income.grossIncome) &&
```
alexose: Hm... I don't think so, since it's a higher-level check, and we already call it [here](https://github.com/department-of-veterans-affairs/healthcare-application/blob/8d49d9b4bdd97a8963208314147ce63febe325bb/src/client/utils/validations.js#L347)

src/client/components/household-information/ChildIncome.jsx
alexose: Done--  This was a little awkward because we don't really have a way to chain multiple validations to different error messages.  It does seem more readable and DRY now though.

akainic:
LGTM! Minus the child income errors, which we'll fix in a separate PR

alexose:
Thanks!

PR #278, opened 2016-11-08
Move location of marital status

akainic:
Moving marital status field out of spouse information section because not all users will now see this section, but marital status is required of all users.

Moving marital status to the demographic page after gender, and then making whether or not we show the spouse page conditional on if they are either Married or Separated.

### DIFF
--- a/src/client/routes.jsx
+++ b/src/client/routes.jsx
@@ -73,7 +73,10 @@ const routes = [
       path="/household-information/financial-disclosure"/>,
   <Route
       component={SpouseInformationSection}
-      depends={{ understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: ['Married', 'Separated'] } }}
+      depends={[
+        { understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: 'Married' } },
+        { understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: 'Separated' } },
+      ]}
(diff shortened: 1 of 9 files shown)


src/client/routes.jsx
```diff
@@ -73,7 +73,7 @@ const routes = [
       path="/household-information/financial-disclosure"/>,
   <Route
       component={SpouseInformationSection}
-      depends={{ understandsFinancialDisclosure: { value: 'Y' } }}
+      depends={{ understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: ['Married', 'Separated'] } }}
```
alexose: The way I'd want to implement this is like so:

```
depends={[
  { understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: 'Married' }},
  { understandsFinancialDisclosure: { value: 'Y' }, maritalStatus: { value: 'Separated' }},
]}
```

akainic:
New commit looks great, thanks @alexose! It would also be great if we could change the nav so the dependent sections don't show up there either.

markolson (approved):

alexose:
Still getting a schema validation error on my end... trying to figure out why

akainic:
@alexose wow! tests passed! does this look good to you? I'm going to create a separate ticket for fixing the nav.

alexose:
👍  Will merge after CI

PR #280, opened 2016-11-10
Fix child income errors

alexose:
After much investigation, I realized that the `fields` array in `uiState` is really only used to mark fields as dirty after the user advances this page.

This is my very silly fix that I think will work, because `children` is marked as dirty after the annual income section as well.

### DIFF
--- a/src/client/reducers/uiState/index.js
+++ b/src/client/reducers/uiState/index.js
@@ -73,7 +73,7 @@ const ui = {
     '/household-information/child-information': {
       complete: false,
       verified: false,
-      fields: ['hasChildrenToReport', 'children']
+      fields: ['hasChildrenToReport']
     },
     '/household-information/annual-income': {
       complete: false,
--- a/src/client/reducers/uiState/index.js
+++ b/src/client/reducers/uiState/index.js
@@ -73,7 +73,7 @@ const ui = {
     '/household-information/child-information': {
       complete: false,
       verified: false,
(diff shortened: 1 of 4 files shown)


akainic:
Unfortunately, this change seems to break the error styling on the children page. If you indicate that you do have children, the children fields expand. But if you leave them all blank and try to click "Continue", you should see error styling on the required fields. With this PR, the error styling no longer appears (though it does still prevent you from continuing, which is good). I think that's because the `dirty` flag isn't being changed to `true` with this change, which means the error styling doesn't appear. 

I'm wondering if we just need to pull out the income fields relevant to the child outside of the `children` array. This will require changing some things in `enrollment-system.js`, but looking at it quickly I think it's doable, and makes more sense given the structure of our application.

alexose:
I see what you mean-- Also, that makes sense given the way this code works.  Darn.

I wanted to avoid changing the data structure, since it adds some complexities to keeping two arrays in sync... still, I think that's probably the right way to do it.

alexose:
I've gone pretty far down the path of trying to populate two arrays, and it's seeming like a lot of work for something pretty kludgy.  Both `Growabletable` and `Fixedtable` both take a single array as input, and there's no reasonable place to clone values from one to the other.

The better approach seems to be improving the `dirtyAllFields` function to support arrays in a nicer way (IMO using `_.get` and property paths).

PR #283, opened 2016-11-16
Include '0' financials in submitted documents

markolson:


src/server/enrollment-system.js
```diff
@@ -241,19 +241,19 @@ function veteranToSpouseInfo(veteran) {
  */
 function resourceToIncomeCollection(resource) {
   const incomeCollection = [];
-  if (resource.grossIncome) {
+  if (resource.grossIncome !== undefined) {
```
jbalboni: This is going to allow nulls through, which might trip us up later. Maybe `!isNaN(resource.grossIncome)`?

jbalboni (approved):
Looks ok to me, though not sure about the `undefined` check.

src/server/enrollment-system.js
markolson: I may be misunderstanding this, but `!isNaN(null) == true`, so null would still get through with that.

jbalboni: Ugh, that's true. In any case, I think we should have a check that checks for null and undefined, to be safe.

markolson: lodash to the rescue.

PR #284, opened 2016-11-17
Require child education expenses

akainic:
We heard that the field from child education expenses needs to be required as the other monetary fields are required.


markolson (approved):

markolson:
This was the only other thing that came up in emails and calls, so big 👍

alexose:
👍

PR #285, opened 2016-11-17
WIP: Add child income fields once on income page

akainic:
Resolves https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/497


markolson:
Yay! I'd add a nice comment about it's split up like it is so that we don't 'fix' or 'simplify' it a few months from now :)

akainic:
@markolson good point! will add. also looks like there are failing tests, so we're digging into that now.

markolson (approved):

PR #286, opened 2016-11-17
Vets json schema package

lihanli:
this schema has 3 extra states: `"AA", "AE", "AP"` which were added to the constants file for education benefits which hca shares. if we don't want this then i can change it


akainic (commented):
LGTM! Do we also want to remove the `schema/validator.js` file?

markolson:
@akainic that's a helper we use in the tests and backend, and should stay for now

markolson (approved):

PR #287, opened 2016-11-18
Production update 111816

akainic:


markolson (approved):
Eyeballing it, everything looks good. I see the important fixes as far as the backend goes.

PR #289, opened 2016-11-18
Remove HCA dependencies

akainic:


markolson (approved):

markolson:
@akainic pushed my commit if you want to review

akainic (commented):
Tested 2 different cases, providing financials and not, and both work for me locally. Looks good!

PR #295, opened 2016-11-21
Revert "Production Deploy 2016-11-21"

akainic:
Reverts department-of-veterans-affairs/healthcare-application#294


akainic:
What's weird is that the reversion only includes that change that should have been made. Looking through all the files updated here (https://github.com/department-of-veterans-affairs/healthcare-application/pull/294) I think they're all right, but it's hard to tell for sure.

knkski (approved):

markolson (approved):

PR #300, opened 2016-11-23
Change package.json to use https: instead of git:

patrickvinograd:
for vets-json-schema fetch, due to AWS requirements.


patrickvinograd:
Tested this by updating package.json on an AWS bake instance and running "npm install", which went through successfully.

Our NAT configuration in AWS does not allow `git:` protocol through.

jkassemi (approved):

PR #301, opened 2016-11-28
Adding basic Jenkinsfile

knkski:


markolson (approved):
For future reference, the plan is that we'll add e2e back once HCA backend is in vets-api, as a daily run.

PR #310, opened 2016-12-14
Update JSON Schema

markolson:


akainic (approved):
LGTM! We don't have to account for this field in enrollment-system.js, right?

markolson:
We don't - by just not doing anything with it at that step, we're not including it in the soap transmission.

PR #312, opened 2016-12-15
Update JSON Schema & add a test to ensure privacyAgreementAccepted=true

markolson:


src/common/veteran.js
```diff
@@ -138,7 +138,8 @@ const blankVeteran = {
   vietnamService: false,
   exposedToRadiation: false,
   radiumTreatments: false,
-  campLejeune: false
+  campLejeune: false,
+  privacyAgreementAccepted: true
```
akainic: Should this be set to `false` first, since this is for an initial blank application?

akainic (commented):
Looks good except for one question!

src/common/veteran.js
markolson: Ok

akainic (approved):
LGTM!

